PR #39, opened 2015-01-05
fixed error string typo (RhBug: 1174712)

jsilhan:


Tojaj:
Hi Jan,
thanks for the patch!
I had to rewrite it, since the current code-base was quite different to the one your patch was written to.

PR #43, opened 2015-02-17
tests/repoconf: Use g_assert_cmpint over libcheck's new symbol

cgwalters:
RHEL7 has check-0.9.9-5.el7 which doesn't have the
ck_assert_uint_eq() symbol.  Since we're linking to GLib, and
it has a (IMO superior) API here, and it's been available
since approximately forever, let's use it.


Tojaj:
Good catch, thanks! I've pushed my version of the patch since use of g_assert_cmpuint instead of g_assert_cmpint is more appropriate.

PR #45, opened 2015-03-16
tests: Pass true(1)/false(0) to ck_assert_msg, fix check.h backcompat

cgwalters:
First; RHEL7.0 does have check-0.9.9 does have ck_assert_msg, so the
previous code was redefining it.

Second, we are passing pointer-sized values as the expression, when an
integer-sized value (really true/false) is expected.  Fix that with
explicit comparisons to NULL.

### DIFF
--- a/tests/fixtures.c
+++ b/tests/fixtures.c
@@ -41,12 +41,12 @@ void
 lr_assert_strv_eq(const gchar * const *strv, ...)
 {
     va_list args;
-    ck_assert_msg(strv, "NULL isn't strv");
+    ck_assert_msg(strv != NULL, "NULL isn't strv");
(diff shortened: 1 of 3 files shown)


Tojaj:
Thanks!

cgwalters:
Actually, this breaks the test suite...and I am quite confused why.  Still debugging it, took me a little bit to spin up on cmake and figure out things like `-DCMAKE_BUILD_TYPE:STRING=Debug` - what I'm present stuck on now is figuring out how to trace through the child processes `srunner` creates in gdb.

cgwalters:
Follow up in https://github.com/Tojaj/librepo/pull/46

PR #46, opened 2015-03-16
tests: Avoid ck_assert_msg() format strings with invalid parameters

cgwalters:
The previous fix to not overwrite ck_assert_msg() means that we
started actually passing format strings.  A pervasive problem
here is that if a check fails, some of the format arguments may
not actually be valid.

For example, if a GError function returns TRUE, the error value will
not be set.

GLib has test utility calls around usage of GError, let's make use of
those.  This drops the format strings, which IMO aren't greatly useful
anyways - if a test fails, it's pretty easy to dive in and see which
option value it's passing, etc.


Tojaj:
Thanks Colin!

PR #48, opened 2015-03-17
CMakeLists.txt: do not check for CXX

giuseppe:
The CXX compiler is not needed, force to check only for C.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


Tojaj:
Thanks!

PR #52, opened 2015-04-08
Support for client certificates

pclifford:
This adds support for setting a client certificate to use when talking to the remote server, and for overriding the path to the CA bundle used to validate the server's certificate.


yavor-atanasov:
That would be amazingly useful to merge. It will have a beneficial effect on the adoption of ``librepo``, ``hawkey`` and ``dnf`` in general as some organisations use client certs. It would also help the resolution of this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1203661

cgwalters:
I'm not a librepo reviewer, but it looks fine to me.

Tojaj:
Looks good! Thanks!

PR #54, opened 2015-04-10
Added a python example for using client certificates with librepo

yavor-atanasov:
This might be evident, but it's always nice to have a distilled example.


Tojaj:
Merged, thanks!

yavor-atanasov:
Cool, thank you Tomas

PR #58, opened 2015-05-18
build: Be compatible with cmake 2.8

cgwalters:
cmake 3.0 adds an optional `LANGUAGES` keyword, which
`cmake-2.8.11-4.el7.x86_64` doesn't support.  Fix it by dropping the
`LANGUAGES` keyword which works in both.

Regression from https://github.com/Tojaj/librepo/pull/48


Tojaj:
Thanks Colin!

PR #59, opened 2015-05-19
downloader: Report last error

cgwalters:
I'm seeing a corrupted RPM in an available repository.  Even when
just using `baseurl=` for the one server, librepo drops into
internal mirorring machinery during downloads and just reports:

```
error: Cannot download bash-4.3.33-2.fc22.x86_64.rpm: All mirrors were tried
```

Which isn't very useful.  Let's cache the last error we hit, so we can
instead say:

```
error: Cannot download bash-4.3.33-2.fc22.x86_64.rpm: All mirrors were tried.  Last error: Downloading successful, but checksum doesn't match. Calculated: 75d062b11bc74d345a6b51620f14c86dc232afa8e0fc6dad28479c0e97cb1d4f(sha256)  Expected: 21b27d1a5b80b933c1e142367624465b5e45163404f85ac2afed01efe260f91e(sha256)
```

(This error message could still be better, but one thing at a time)

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -188,6 +188,9 @@ typedef struct {
     GSList *running_transfers; /*!<
         List of running transfers (list of pointer to LrTarget structures) */
 
+    GError *last_transfer_error; /*!<
+        The error encountered from the last mirror we tried */
+
 } LrDownload;
 
 /** Schema of structures as used in downloader module:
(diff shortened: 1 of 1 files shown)


hughsie:
I've wanted this for so long!

Tojaj:
Hm, this is a duplicating already existing functionality (Mirror failure callback - LrMirrorFailureCb).

The callback could be set for each package target and it reports each mirror failure. (You have to use lr_packagetarget_new_v2() and higher to be able to set the callback)

E.g. This python code:

    #!/usr/bin/env python

    import librepo


    urls = [
        "http://foo.bar/",
        "http://redhat.com/",
        "ftp://mirror.vutbr.cz/fedora/releases/21/Server/x86_64/os/",
    ]

    def mirrorfailurecb(_, msg, url):
        print "Download of %s failed: %s" % (url, msg)

    h = librepo.Handle()
    h.setopt(librepo.LRO_URLS, urls)
    h.setopt(librepo.LRO_REPOTYPE, librepo.LR_YUMREPO)

    targets = []
    target = librepo.PackageTarget("Packages/a/abrt-2.3.0-3.fc21.x86_64.rpm",
                                   handle=h,
                                   mirrorfailurecb=mirrorfailurecb)
    targets.append(target)

    librepo.download_packages(targets)

Returns:

    Download of http://foo.bar/Packages/a/abrt-2.3.0-3.fc21.x86_64.rpm failed: Curl error (6): Couldn't resolve host name for http://foo.bar/Packages/a/abrt-2.3.0-3.fc21.x86_64.rpm [Could not resolve host: foo.bar]
    Download of http://www.redhat.com/Packages/a/abrt-2.3.0-3.fc21.x86_64.rpm failed: Status code: 404 for http://www.redhat.com/Packages/a/abrt-2.3.0-3.fc21.x86_64.rpm


Note: I don't like the approach where only the last error is reported, because last error could be misleading.
But if you both want this, I will do the merge.

cgwalters:
I'll look at using the mirrorfailurecb in libhif; thanks for the detailed reply.  If I hit any problems I'll open a different issue.  Thanks!

PR #61, opened 2015-06-17
Support redefining %{gitrev} RPM macro from command line.

rholy:
This together with the following change to the spec file:

```diff
diff --git a/librepo.spec b/librepo.spec
index d5582db..0884735 100644
--- a/librepo.spec
+++ b/librepo.spec
@@ -1,4 +1,4 @@
-%global gitrev d9bed0d
+%{!?gitrev: %global gitrev d9bed0d}
 # gitrev is output of: git rev-parse --short HEAD
 
 %if 0%{?rhel} != 0 && 0%{?rhel} <= 7
```

allows one to build librepo RPMs in Mock more easily than running `make_rpm.sh` in Mock (which requires copying files to and from the chroot and installing all the dependencies):

```bash
GITREV=`git rev-parse --short HEAD`
utils/make_tarball.sh "$GITREV"
mock --resultdir=. --define="gitrev $GITREV" --buildsrpm --spec librepo.spec --sources .
mock --resultdir=. --define="gitrev $GITREV" *.src.rpm
```

Also `make_rpm.sh` can get rid of the `sed` but I guess you still want to update the spec file with the new GITREV.

It will help us to build librepo nightly builds.

Let me know if there is any problem/ugliness.

### DIFF
--- a/utils/make_rpm.sh
+++ b/utils/make_rpm.sh
@@ -45,9 +45,8 @@ if [ ! $? == "0" ]; then
     exit 1
 fi
 
-# Copy via sed
-sed -i "s/%global gitrev .*/%global gitrev $GITREV/g" "$PREFIX/$PACKAGE.spec"
-sed "s/%global gitrev .*/%global gitrev $GITREV/g" "$PREFIX/$PACKAGE.spec" > "$RPMBUILD_DIR/SPECS/$PACKAGE.spec"
+sed -i "s/%{\!?gitrev: %global gitrev .*/%{\!?gitrev: %global gitrev $GITREV}/g" "$PREFIX/$PACKAGE.spec"
+cp "$PREFIX/$PACKAGE.spec" "$RPMBUILD_DIR/SPECS/"
 if [ ! $? == "0" ]; then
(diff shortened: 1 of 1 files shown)


rholy:
So, we no longer build librepo in Mock so it's not such a pain. We now build just SRPMs and upload them to Copr. But unless you provide [a script for building a SRPM](https://github.com/Tojaj/librepo/issues/70) (which also modifies GITREV), this change would still simplify our script (we could get rid of the `sed`).

rholy:
Thank you. I don't see change in the spec file so I hope you don't forget to apply the patch above. Otherwise, this would just broke your workflow.

Tojaj:
I will change the spec in Fedora, but before that, I want to apply few other patches and release a new minor version of Librepo. Stay tuned.

Tojaj:
Rawhide build is ready => New spec is available in distgit (just for rawhide so far):
koji.fedoraproject.org/koji/taskinfo?taskID=11226650

rholy:
I just wanted to be sure that I did not break your workflow. Thanks again!

rholy:
Actually, we build RPM only from master so **we** don't need any spec update in the other branches. But of course, `make_rpm` wouldn't work for **you** unless you do the change...

PR #65, opened 2015-07-22
Link against libssl

bgamari:
checksum.c clearly depends upon OpenSSL's EVP implementation. Previously this would result in undefined symbols at runtime.


Conan-Kudo:
This fixed my build for Mageia 5/Cauldron 👍

cgwalters:
Is your build still fixed with only `-lcrypto`?

cgwalters:
@Conan-Kudo :arrow_up:

Conan-Kudo:
@cgwalters Looking into it. I need to make some adjustments to use libcrypto instead of this patch.

cgwalters:
The reason I'm talking about this is `libcrypto` is the sane part of OpenSSL, and there's really reason for librepo itself to directly need TLS - that comes from curl.

Tojaj:
Fixed, could someone of you test this on a system where the previous builds failed?

Conan-Kudo:
The fix to use libcrypto works perfectly on Mageia 5 and Cauldron.

Tojaj:
@Conan-Kudo :+1:

Issue #70, opened 2015-09-08
Librepo cannot be build in Copr (and potentially other environments)


rholy:
Apparently, the Copr environment does not support extended file attributes, thus the `test_download_packages_with_resume_02` test fails there, thus the RPM cannot be build there.

Here is one of the failing builds: https://copr-be.cloud.fedoraproject.org/results/rholy/test/fedora-rawhide-x86_64/00113768-librepo/
```
[...]
2: ======================================================================
2: ERROR: test_download_packages_with_resume_02 (tests.test_yum_package_downloading.TestCaseYumPackagesDownloading)
2: ----------------------------------------------------------------------
2: Traceback (most recent call last):
2:   File "/builddir/build/BUILD/librepo/tests/python/tests/test_yum_package_downloading.py", line 688, in test_download_packages_with_resume_02
2:     xattr.setxattr(pkg.local_path, "user.Librepo.DownloadInProgress", "")
2: IOError: [Errno 95] Operation not supported
[...]
```

(the other solution would be to ask Copr maintainers to allow xattrs but I believe that librepo should be buildable even if xattrs are not allowed)

rholy:
BTW, once fixed, librepo nightly builds will be available in [rpmsoftwaremanagement/librepo-nightly-fedora-rawhide](https://copr.fedoraproject.org/coprs/rpmsoftwaremanagement/librepo-nightly-fedora-rawhide/), [rpmsoftwaremanagement/librepo-nightly-fedora-23](https://copr.fedoraproject.org/coprs/rpmsoftwaremanagement/librepo-nightly-fedora-23/) and [rpmsoftwaremanagement/librepo-nightly-fedora-22 ](https://copr.fedoraproject.org/coprs/rpmsoftwaremanagement/librepo-nightly-fedora-22/)

Tojaj:
Closing as fixed by your patch, thanks!

rholy:
Cool, thank you. There are already some librepo builds available in that Copr.

PR #71, opened 2015-09-10
tests: Skip the relevant tests if extended attributes are not supported

rholy:
Issue #70

Please review this carefully. I don't know your coding conventions and I don't know whether you really want to skip the test. For me, it seems correct. I may also misunderstood the test.

Also note that `Unittest.skipTest` is supported on Python 2.7 and Python 3.1+ so if you support older Pythons, I need to find a different solution.

Otherwise, I'd appreciate if you could merge it as soon as possible.


Tojaj:
Thanks Radek!

PR #72, opened 2015-09-21
Switch `command -v` for `which` in nosetest check

Conan-Kudo:
Some distributions (such as Mageia) do not provide `/usr/bin/command`, and instead rely on it being an internal command in bash. However, that breaks the test in CMake for versioned nosetests binaries.

To correct this, I switched it to use `/usr/bin/which` to enable the same test for versioned nosetests binaries.


Tojaj:
At first I wanted to use ``bash command -v ...`` instead. Just to avoid ``which`` [1], but in the end I decided to merge the patch as is.
May be changed in future if any problems with ``which`` should appear.

[1] http://stackoverflow.com/a/677212/2779058

Conan-Kudo:
@Tojaj I tried `bash command -v` at first, but that didn't work...

Tojaj:
Ok, thanks for info

PR #75, opened 2015-11-24
New test: always try to download from the fastest mirror

atodorov:
When a package fails to download, librepo moves to the next mirror;
after a successfull download librepo should continue downloading
the remaining packages from the first mirror. In yum this was not
the case and every download failure was causing the usage of slower
mirrors even when the fastest ones had the remaining packages. This
test verifies this is not the case with librepo.

Note: I'm verifying the order in which packages failed to download from the mirrors because there's no way to verify that packages came from a particular mirror. If you change the order of the URLs in the metalink file the test will fail.


Tojaj:
Thank you!

PR #77, opened 2016-01-12
Tolerate HTTP proxies with (conformant) heuristic caching

sourcejedi:
> dnf failed sometimes when using a Polipo proxy.  repomd.xml was cached
> with out-of-data contents, causing download failure.  To avoid this,
> dnf should force revalidation ("Cache-Control: max-age=0") on repomd.xml.

> https://bugzilla.redhat.com/show_bug.cgi?id=1297762

I tested this on my own machine.  It seemed to send the desired Cache-Control header, and completed `dnf update`.  It was able to connect through my proxy mirror, as well as directly using the normal metalink.

I haven't tried to reproduce the failure condition.  I do want to keep using Polipo, and hopefully I'll use `dnf` enough to notice if it's still breaking.  It took at most 10 days to see a failure the first time.

DISCLAIMER I'm still slightly anxious.  I don't know how to assess the risk that this will break for some _other_ proxy!  Even though I can't imagine how it would break.

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -147,6 +147,8 @@ typedef struct {
         range was downloaded, it is TRUE. Otherwise FALSE. */
     LrCbReturnCode cb_return_code; /*!<
         Last cb return code. */
+    struct curl_slist *curl_rqheaders; /*!<
+        Extra headers for request. */
 } LrTarget;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


sourcejedi:
Thanks for review!  I'm looking at this now.  I like the simplification to `g_assert()` in your last comment.  A disadvantage of adding new `g_assert()` checks is the number of missing checks in the function...  I'm thinking of using `||=`.

PR #82, opened 2016-02-08
Caching proxy fix

sourcejedi:
> dnf failed sometimes when using a Polipo proxy.  repomd.xml was cached
> with out-of-data contents, causing download failure.  To avoid this,
> dnf should force revalidation ("Cache-Control: max-age=0") on repomd.xml.

> https://bugzilla.redhat.com/show_bug.cgi?id=1297762

I've seen the failure twice now.  Updating from a machine with the fixed librepo appeared to successfully flush the stale files from the caching proxy.

Thanks to @cgwalters for reviewing PR #77, my first attempt at this!

The skipped curl error checks I noticed in the modified function have been added in one final patch, because we weren't 100% sure about it.  Looks tolerable to me, though maybe only because I've been learning how to handle errors in Rust :).

downloader.c and elsewhere don't use `g_assert()`, so I've stuck with `assert()` for now.  Commit gives a justification for using `assert()`.  That doesn't quite apply to the `curl_multi_add_handle()` commit.  But I see `assert(!err || *err == NULL)` elsewhere, and I have to stop digging / yak shaving at some point...

<small>Further to my original comment, `||=` is not a thing in C :).  I agree it would be good to look at `__attribute__((cleanup))` for cleanup... and that this fix is not the right place for it.  (I only cared about code quality here, because I want _my_ new code to be both correct and consistent with surrounding code :-P).</small>

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -974,8 +974,9 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
         g_debug("%s: Used offset for download resume: %"G_GINT64_FORMAT,
                 __func__, used_offset);
 
-        curl_easy_setopt(h, CURLOPT_RESUME_FROM_LARGE,
-                         (curl_off_t) used_offset);
+        c_rc = curl_easy_setopt(h, CURLOPT_RESUME_FROM_LARGE,
+                                (curl_off_t) used_offset);
+        assert(c_rc == CURLE_OK);
     }
 
     // Add librepo extended attribute to the file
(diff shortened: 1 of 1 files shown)


cgwalters:
All these patches LGTM.

Tojaj:
@sourcejedi 
Hi Alan,
thank you very much for the patches!

They looks good, but the main of them **Tolerate HTTP proxies with (conformant) heuristic caching** couldn't be merged. I see few problems right now:

* It breaks functionality of LRO_HTTPHEADER option [1] because it destructively overrides it.
* Not only repomd.xml needs fresh data, also metalink.xml, mirrorlist, etc. need it.
* **Cache-Control** header may not be sufficient. In order to support also HTTP/1.0 proxies, we should send also **Pragma: no-cache** [2] 

I'll take a look at this and will let you know.

[1] http://rpm-software-management.github.io/librepo/lib.html#librepo.LRO_HTTPHEADER
[2] https://curl.haxx.se/mail/lib-2003-10/0137.html

Tojaj:
I've wrote patches for no cache HTTP headers:
* 473b92f964f33734555ca9f7f5093cd2ab17d98e
* 6f669c0afead047068ec03af8783483a6aa77eea
* 0141392e9f8f7ad9135b62d92d75c4cb65d57191
* a47370b370ba5ae95fb6220d09529a9a42b90ec5

Tojaj:
@sourcejedi Could you please update the PR and exclude the "Tolerate HTTP proxies with (conformant) heuristic caching" patch?

sourcejedi:
Thanks!  I agree with all your reasoning.  I will repay you with a rebase of these patches... soon :).

sourcejedi:
Ok, I've rebased it.  It includes the checks for out of memory on `curl_slist_append()`, to your version of the option-setting code.

Should probably have put this in the commit, but - `assert()` looks fine to check the result `CURLOPT_HTTPHEADER`, even though we're including user-specified headers, because the doc  says it only fails if curl is built without HTTP.  (And hopefully we get a helpful error message from setting a HTTP URL, before that point!).

Tojaj:
Looks good! Thanks!

PR #84, opened 2016-03-16
yum: Fix a memory leak in signature verification

cgwalters:
This was caught by the libhif valgrind tests.

### DIFF
--- a/librepo/yum.c
+++ b/librepo/yum.c
@@ -875,6 +875,7 @@ lr_yum_download_remote(LrHandle *handle, LrResult *result, GError **err)
                                              path,
                                              handle->gnupghomedir,
                                              &tmp_err);
+                lr_free(signature);
                 if (!ret) {
                     g_debug("%s: GPG signature verification failed: %s",
                             __func__, tmp_err->message);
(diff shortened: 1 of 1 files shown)


cgwalters:
Also, any objections to converting the whole codebase to use `g_autofree`?   If you have concerns about requiring a newer glib, simply carrying a backport of `g_autofree` would be easy.

Tojaj:
@cgwalters Merged, thank you Colin!

The ``g_autofree`` looks really powerful, I didn't know that macro.
Use of it would require more refactoring, because Librepo uses also classical malloc and calloc function and not just glib's alternatives. Also the backport would be absolutely necessary, because Glib 2.44 is still quite bleeding-edge.

cgwalters:
In modern glib we basically require `g_malloc = malloc` since the gobject initialization was moved to `__attribute__ ((constructor))`.

Okay, I might take a look at taking the first steps for `g_autofree` here.

If you look at codebases like systemd, ostree, increasingly NetworkManager, now libhif, and others, the code readability is dramatically improved.  See https://plus.google.com/+LennartPoetteringTheOneAndOnly/posts/jHPAdLJiw23

PR #87, opened 2016-07-25
add yum record substitution mechanism

mluscon:
This could be used from python via `handle.yumslist = [('group', 'group_gz')]` which effectively means - download `group` if `group_gz` is not listed in metadata.

I mostly reused the LrUrlVars handling in the python part that is unfortunately too tied with variables substitution. We should probably rename it to something more general for a better readability of this code.

@Tojaj, please let know whether you agree with this approach. If so, I will write some doc and tests.


librepo/yum.c
```diff
@@ -152,9 +162,29 @@ lr_yum_repomd_record_enabled(LrHandle *handle, const char *type)
                 return TRUE;
             x++;
         }
+        // Substitution check
+        if (handle->yumslist) {
+            LrUrlVars* elem = handle->yumslist;
+            while (TRUE) {
```
Tojaj: Instead of this two lines (167-168) + lines 181-184, I would prefer:

    for (GSList *elem = repomd->records; elem; elem = g_slist_next(elem)) {

which does exactly the same thing, but it's more compact and readable

librepo/yum.c
```diff
@@ -152,9 +162,29 @@ lr_yum_repomd_record_enabled(LrHandle *handle, const char *type)
...
+                LrVar* subs = elem->data;
+                if (!strcmp(subs->var, type)) {
```
Tojaj: Maybe use g_strcmp0() as it's more bullet-prof (NULL won't cause SIGSEGV)

https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0

librepo/yum.c
```diff
@@ -152,9 +162,29 @@ lr_yum_repomd_record_enabled(LrHandle *handle, const char *type)
...
+                    char *orig = subs->val;
+                    int y = 0;
+                    while (handle->yumdlist[y]) {
+                        if (!strcmp(orig, handle->yumdlist[y]) &&
```
Tojaj: Here you can also save lines 172, 173, 177 by:

    for (guint i = 0; handle->yumdlist[i]; i++) {

librepo/yum.c
```diff
@@ -131,8 +131,18 @@ lr_yum_repo_update(LrYumRepo *repo, const char *type, const char *path)
 
 /* main bussines logic */
 
+gint
+compare_records(gconstpointer a, gconstpointer b)
+{
+    GSList* elem = (GSList*) a;
+    LrYumRepoMdRecord* yum_record = (LrYumRepoMdRecord*) elem;
```
Tojaj: Please access explicitly the ``data`` element of GSList here:

    elem->data;

See: https://developer.gnome.org/glib/stable/glib-Singly-Linked-Lists.html#GSList

You can directly merge lines 137-138 by:

    LrYumRepoMdRecord* yum_record = (GSList*) a->data;

librepo/yum.c
```diff
@@ -131,8 +131,18 @@ lr_yum_repo_update(LrYumRepo *repo, const char *type, const char *path)
...
+    char *type1 = (char *) yum_record->type;
+    char *type2 = (char *) b;
+    return strcmp(type1, type2);
```
Tojaj: Maybe use g_strcmp0() as it's more bullet-prof (NULL won't cause SIGSEGV)

https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0

Tojaj:
Thank you for the PR, I added few comments

Tojaj:
One more note about order: ``[('group', 'group_gz')]``
I would expect that this says: group could be replaced by group_gz but your semantic is opposite.
Originally url vars works also that way 1st is var 2nd is replacement

mluscon:
Ok, I incorporated requested changes.

librepo/yum.c
```diff
@@ -152,9 +161,23 @@ lr_yum_repomd_record_enabled(LrHandle *handle, const char *type)
                 return TRUE;
             x++;
         }
+        // Substitution check
+        if (handle->yumslist) {
+            for (GSList *elem = handle->yumslist; elem; elem = g_slist_next(elem)) {
+                LrVar* subs = elem->data;
+                if (!g_strcmp0(subs->val, type)) {
+                    char *orig = subs->var;
+                    for (guint i = 0; handle->yumdlist[i]; i++) {
```
Tojaj: I see that you iterate over ``yumdlist`` and not over ``yumslist`` which is guaranteed to be not NULL in this part of code. For that reason (yumdlist may be NULL here), there should be:

    for (guint i = 0; handle->yumdlist && handle->yumdlist[i]; i++) {

I.e. make sure that ``handle->yumdlist`` is not NULL before to try access its elements via index

Tojaj:
Nice update, thank you, but I've noticed one more possible issue.
Btw. would it be possible to write at least a simple test (in python) for this?

Just set yumdlist to "primary" and "foo" and set yumslist to [("foo", "other")] and check if the downloaded repo has primary and other repodata present.
(There are already bunch of unittests, just copy one test function and modify it to match the use case)

librepo/yum.c
mluscon: This whole thing is inside `if (handle->yumdlist)` so there is no need to check it again.

mluscon:
test++

PR #88, opened 2016-09-07
util: lr_prepend_url_protocol: don't resolve relative repo URLs using cwd

sourcejedi:
It won't go well when those repo URLs come from config files.

Users specifying repo URLs as run-time options shouldn't expect special
treatment.  They can easily use `$(pwd)`

`test_download_packages_with_baseurl` relied on using "." as a non-existent
repo, for some reason.  Fix it to use config.BADURL.  Copied+pasted from
`test_download_package_with_baseurl`.

### DIFF
--- a/librepo/handle.c
+++ b/librepo/handle.c
@@ -796,12 +796,9 @@ lr_handle_prepare_urls(LrHandle *handle, GError **err)
         _cleanup_free_ gchar *final_url = NULL;
 
         // Make sure that url has protocol specified
-        final_url = lr_prepend_url_protocol(url);
-        if (!final_url) {
-            g_set_error(err, LR_HANDLE_ERROR, LRE_BADURL,
-                        "Cannot resolve path for: \"%s\"", url);
+        final_url = lr_prepend_url_protocol(url, err);
+        if (!final_url)
             return FALSE;
-        }
 
         // Append the url into internal list of urls specified by LRO_URLS
         handle->urls_mirrors = lr_lrmirrorlist_append_url(
@@ -870,8 +867,8 @@ lr_handle_prepare_mirrorlist(LrHandle *handle, gchar *localpath, GError **err)
             return FALSE;
         }
 
-        url = lr_prepend_url_protocol(handle->mirrorlisturl);
-        if (!download_non_cached_url(handle, url, fd, err)) {
+        url = lr_prepend_url_protocol(handle->mirrorlisturl, err);
+        if (!url || !download_non_cached_url(handle, url, fd, err)) {
             close(fd);
             return FALSE;
         }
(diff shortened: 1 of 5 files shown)


sourcejedi:
Thanks for looking at this :).  You said the tests now pass for you with PR #89.  Are they passing for this one as well now (PR#88), or is there something I need to look at?

Conan-Kudo:
@sourcejedi They are broken for this PR, but the other one is fine.

sourcejedi:
I see it.  Thanks again, I have squashed the following into the commit:

```
diff --git a/tests/python/tests/test_yum_package_downloading.py b/tests/python/tests/test_yum_package_downloading.py
index 18df83a..8020129 100644
--- a/tests/python/tests/test_yum_package_downloading.py
+++ b/tests/python/tests/test_yum_package_downloading.py
@@ -413,7 +413,7 @@ class TestCaseYumPackagesDownloading(TestCaseWithFlask):
         pkgs.append(librepo.PackageTarget(config.PACKAGE_01_01,
                                           handle=h,
                                           dest=self.tmpdir,
-                                          base_url=url))
+                                          base_url=baseurl))
 
         librepo.download_packages(pkgs)

```

kalev:
This changes lr_prepend_url_protocol() signature and as such the whole library public API -- I wonder if an API change is worth doing it at this point? Maybe just do the checks on the caller side for now instead of adding an error param?

It would need a soname bump if going forward with this.

sourcejedi:
Ugh, thanks.  GitHub search finds one use of `lr_prepend_url_protocol()` in libdnf, where the behavioral fix is also relevant.  Google also finds in passing that `lr_prepend_url_protocol()` was used by libhif.

I'm not inclined to add the check `!strstr(path, "://") && path[0] != '/'` to callers (and leave the dead code).  I think they should use a fixed version of this function.

However it is starting to look like this is not worth the effort, in the absence of any user complaints.

All I can argue for the soname bump, is that it surprises me to realize `util.h` is included in `librepo.h`, it's not treated similar to the `_internal.h` files (some of which do include extern function declarations), and there are _no_ internal utilities. Is this intentional? Have `util.h` functions been designed as public apis? What is the purpose of `lr_malloc()` as an API (the behaviour of which is not documented)?

Bumping the soname would be an opportunity to exclude other utilities from public headers.  I understand this doesn't stop people using them if they declare them themself :).  However a revised `lr_prepend_url_protocol()` should probably still be public, and I don't know which of the other utilities have been found useful by users of librepo.

kalev:
I don't really have a strong opinion either way. Can you just amend this to bump soname as well if you want to keep it like this?

@Tojaj Can you look into merging this and the other pending pull request, please? They look like very nice fixes to me.

PR #89, opened 2016-09-07
Improve error handling, cleanup

sourcejedi:
@cgwalters suggested the cleanup attribute could be used to make sure we
don't leak resources when prepare_next_transfer() fails, as this was overlooked in
the past.
#77 (comment)

It took a while to untangle to my satisfaction, but I think it helps.  It turns out we'd still overlooked closing the one fd in there :).

At the same I noticed a "few" other error paths which could be improved.  Enjoy!

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -2102,7 +2102,7 @@ check_transfer_statuses(LrDownload *dd, GError **err)
 
     while ((msg = curl_multi_info_read(dd->multi_handle, &msgs_in_queue))) {
         LrTarget *target = NULL;
-        char *effective_url = NULL;
+        _cleanup_free_ char *effective_url = NULL;
         int fd;
         gboolean matches = TRUE;
         GError *transfer_err = NULL;
@@ -2395,8 +2395,6 @@ transfer_error:
             lr_downloadtarget_set_effectiveurl(target->target,
                                                effective_url);
         }
-
-        lr_free(effective_url);
 
         if (fail_fast_error) {
             // Interrupt whole downloading
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -1321,13 +1321,57 @@ check_zck(LrTarget *target, GError **err)
 }
 #endif /* WITH_ZCHUNK */
 
-/** Prepares next transfer
+/** Open the file to write to
+ */
+static FILE*
+open_target_file(LrTarget *target, GError **err)
+{
+    int fd;
+    FILE *f;
+
+    if (target->target->fd != -1) {
+        // Use supplied filedescriptor
+        fd = dup(target->target->fd);
+        if (fd == -1) {
+            g_set_error(err, LR_DOWNLOADER_ERROR, LRE_IO,
+                        "dup(%d) failed: %s",
+                        target->target->fd, g_strerror(errno));
+           return NULL;
+        }
+    } else {
+        // Use supplied filename
+        int open_flags = O_CREAT|O_TRUNC|O_RDWR;
+        if (target->resume || target->target->is_zchunk)
+            open_flags &= ~O_TRUNC;
+
+        fd = open(target->target->fn, open_flags, 0666);
+        if (fd == -1) {
+            g_set_error(err, LR_DOWNLOADER_ERROR, LRE_IO,
+                        "Cannot open %s: %s",
+                        target->target->fn, g_strerror(errno));
+            return NULL;
+        }
+    }
+
+    f = fdopen(fd, "w+b");
+    if (!f) {
+        close(fd);
+        g_set_error(err, LR_DOWNLOADER_ERROR, LRE_IO,
+                    "fdopen(%d) failed: %s",
+                    fd, g_strerror(errno));
+        return NULL;
+    }
+
+    return f;
+}
+
+/** Prepare next transfer
  */
 static gboolean
 prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
 {
     LrTarget *target;
-    char *full_url = NULL;
+    _cleanup_free_ char *full_url = NULL;
     LrProtocol protocol = LR_PROTOCOL_OTHER;
     gboolean ret;
     GError *tmp_err = NULL;
@@ -1361,8 +1405,9 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
         // Something went wrong
         g_set_error(err, LR_DOWNLOADER_ERROR, LRE_CURL,
                     "curl_easy_duphandle() call failed");
-        return FALSE;
+        goto fail;
     }
+    target->curl_handle = h;
 
     // Set URL
     c_rc = curl_easy_setopt(h, CURLOPT_URL, full_url);
@@ -1370,9 +1415,7 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
         g_set_error(err, LR_DOWNLOADER_ERROR, LRE_CURL,
                     "curl_easy_setopt(h, CURLOPT_URL, %s) failed: %s",
                     full_url, curl_easy_strerror(c_rc));
-        lr_free(full_url);
-        curl_easy_cleanup(h);
-        return FALSE;
+        goto fail;
     }
 
     // Set error buffer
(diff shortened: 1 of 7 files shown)


librepo/downloader.c
```diff
@@ -1329,7 +1330,7 @@ check_finished_transfer_checksum(int fd,
                                               &calculated,
                                               err);
         if (!ret)
```
Tojaj: IMHO, this won't work properly.

The ``ret`` which is tested here is a ``ret`` var defined in this context (context of the ``for`` loop during call of ``lr_checksum_fd_compare``). The newly defined ``ret`` in the main context of ``check_finished_transfer_checksum`` function will still be ``TRUE`` (it won't be set to FALSE) and after the control jump to the ``cleanup`` it will return ``TRUE`` and ``GError`` instead of ``FALSE`` and ``GError``.

Tojaj: To fix this, just replace:

    gboolean ret = lr_checksum_fd_compare(chksum->type,

with:

    ret = lr_checksum_fd_compare(chksum->type,

ignatenkobrain:
ping @sourcejedi ?

sourcejedi:
Sorry for the wait after you'd reviewed this. I could see you were right about the usage of `ret`. IIRC, I made a mistake when rebasing that change.

As a result, I wanted to check things over again.  (Which is a bit of a job, and I never got round to it).  I should be able to get it sorted this weekend.

So far I've read over all the other changes to `downloader.c`, and I'm happy with them.  (Also I've noticed it would be warned about by `CFLAGS=-Wshadow`, but this is not included in `-Wall`. Apparently -Wshadow is not included in -Wextra either.  -Wshadow also detects one other, harmless "shadowed" variable).

sourcejedi:
Done!

ignatenkobrain:
Jenkins, retest this please

Conan-Kudo:
Does any of this still apply anymore? Or should this just be closed?

sourcejedi:
Sure, everything except downloader.c applies.  The addition of `#ifdef WITH_ZCHUNK` in prepare_next_transfer() felt a bit unfortunate - adding yet another code path in there :-), that does `return TRUE`.  But I think we just keep that specific path how it is, and the merge conflict will resolve fine.

I might do this twice & link you a history that does a merge instead of a rebase. I think my tools are better at merging, but I assume you'll want a full rebase.

sourcejedi:
OK, done.  I would quite like to see at least up to the sixth commit applied.  That one was the motivation for this PR.

I think the new ZCHUNK code got the cleanup correct, it just added yet another duplication of the cleanup lines in prepare_next_transfer().  This PR removes the duplicated cleanup in the `return FALSE` branch.  The `return TRUE` branch is more unique, so I left it where it was (and adjusted it to fit my changes).

The merge history version is at https://github.com/rpm-software-management/librepo/compare/master...sourcejedi:use_cleanup_merge?expand=1 .  GitHub doesn't really understand merge commits though.  `gitk` highlights the conflict resolution lines nicely in black (prefixed with "++").  I've copied & pasted them.

3 conflicts:

```
 +    } else {
 +        // Use supplied filename
 +        int open_flags = O_CREAT|O_TRUNC|O_RDWR;
-         if (target->resume)
++        if (target->resume || target->target->is_zchunk)
 +            open_flags &= ~O_TRUNC;
```

```
@@@ -937,8 -1431,32 +1435,34 @@@
      target->writecb_recieved = 0;
      target->writecb_required_range_written = FALSE;
  
+     #ifdef WITH_ZCHUNK
+     // If file is zchunk, prep it
+     if(target->target->is_zchunk && !check_zck(target, &tmp_err)) {
+         g_set_error(err, LR_DOWNLOADER_ERROR, LRE_ZCK,
+                     "Unable to initialize zchunk file %s: %s",
+                     target->target->path,
+                     tmp_err->message);
 -        curl_easy_cleanup(h);
 -        return FALSE;
++        goto fail;
+     }
+ 
+     // If zchunk is finished, we're done
+     if(target->target->is_zchunk &&
+        target->zck_state == LR_ZCK_DL_FINISHED) {
+         g_debug("%s: Target already fully downloaded: %s", __func__, target->target->path);
+         target->state = LR_DS_FINISHED;
 -        curl_easy_cleanup(h);
++        curl_easy_cleanup(target->handle);
++        target->handle = NULL;
+         g_free(target->headercb_interrupt_reason);
+         target->headercb_interrupt_reason = NULL;
+         fclose(target->f);
+         target->f = NULL;
+         lr_downloadtarget_set_error(target->target, LRE_OK, NULL);
+         return TRUE;
+     }
+     # endif /* WITH_ZCHUNK */
+ 
 +    int fd = fileno(target->f);
 +
      // Allow resume only for files that were originally being
      // downloaded by librepo
      if (target->resume && !has_librepo_xattr(fd)) {
```

```
+         #endif /* WITH_ZCHUNK */
 -            ret = check_finished_trasfer_checksum(fd,
++            ret = check_finished_transfer_checksum(fd,
```

Fourthly, I added a code style change.  This shows up as a separate commit in the merge history version: https://github.com/rpm-software-management/librepo/commit/8955a36fe4b219

Fifthly, the compiler nagged me into adding [another commit on the end](https://github.com/rpm-software-management/librepo/pull/89/commits/ba08d443b5a02420b32341145021f14cd25b1a09). "downloader prepare_next_transfer: remove unused variable".

sourcejedi:
F***.  Wait, somehow I *still* didn't fix my mistake that Tojaj pointed, out in check_finished_transfer_checksum().

librepo/downloader.c
sourcejedi: Indeed.  This should finally be fixed now, as per your instructions.

sourcejedi:
Ah, I see what happened.  I had left the "bad" commit, but added a later commit to fix it.  Now I rebased to fix the bad commit.  (The later fix commit was automatically dropped, because it was now empty).  The resulting tree is the same, but the history has been rewritten.

And that's for the best.

Since I had to rebase to master anyway, there was no point leaving the old history intact.  (Again, you can look at the branch `use_cleanup_merge` in my repo.  That still has the "real" history, and uses a merge commit to generate the exact same tree).

Conan-Kudo:
@sourcejedi No, this is fine.

But to answer your questions about librepo: it is used by _everything_ involved with package management now, and a number of tools use it independently to fetch repodata for their own purposes. It was created as a separate library to facilitate those use-cases while not making the DNF API itself host those.

Conan-Kudo:
@dmach Could you please have someone take a look at this and merge it if it's all gravy?

jrohel:
@sourcejedi Thanks for PR. I started review.
But the librepo is not compilable with your PR.
librepo/librepo/downloader.c:1439:57: error: ‘tmp_err’ undeclared (first use in this function)
Problem with ZCHUNK. Please, can you do a small fix?

sourcejedi:
Heh. I thought that was odd. This seems to fix it ^.

Thanks for looking at this.

librepo/downloader.c
```diff
@@ -1437,27 +1436,30 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
...
+            target->state = LR_DS_FINISHED;
+            curl_easy_cleanup(target->handle);
+            target->handle = NULL;
+            g_free(target->headercb_interrupt_reason);
+            target->headercb_interrupt_reason = NULL;
+            fclose(target->f);
+            target->f = NULL;
+            lr_downloadtarget_set_error(target->target, LRE_OK, NULL);
+            return TRUE;
+	}
```
jrohel: Please be consistent. Use 8 spaces instead of tab.

jrohel:
@sourcejedi I finished review. Thanks for your work.
I have only one cosmetics request. Please change one line in the last commit. Use 8 spaces instead of tab.

sourcejedi:
Ah, I should have configured my editor!  Fixed.

Thanks again.  I will be happy to see this merged.  On the bright side, this pull request has been reviewed by a number of people now, I can hope there are no bugs left in it :-).

jrohel:
@sourcejedi Probably you forgot  commit the fix, before you made "force push". I see the old version with tab character.

sourcejedi:
Right.  I've committed and pushed now.  Thanks for noticing the whitespace error (twice!).

jrohel:
@sourcejedi  Thank you for your contribution and patience. I'm going to merge your PR.

PR #94, opened 2016-12-20
Don't resolve relative repo URLs using cwd

sourcejedi:
This is a second attempt at #88.

This version avoids causing segfaults due to changing a public ABI without bumping soname.  It also avoids breaking the public API (requiring a flag day in users of the specific API).

Instead, a new function is added.  The existing one is rewritten as a wrapper to the old one, and still passes the tests.  The existing function is marked as deprecated, for maximum clarity.

There are known users of the API who will see deprecation warnings.  To remove the warnings, they will need a change which depends on this version of librepo.  This pull-request does not co-ordinate any such changes.

### DIFF
--- a/librepo/handle.c
+++ b/librepo/handle.c
@@ -796,12 +796,9 @@ lr_handle_prepare_urls(LrHandle *handle, GError **err)
         _cleanup_free_ gchar *final_url = NULL;
 
         // Make sure that url has protocol specified
-        final_url = lr_prepend_url_protocol(url);
-        if (!final_url) {
-            g_set_error(err, LR_HANDLE_ERROR, LRE_BADURL,
-                        "Cannot resolve path for: \"%s\"", url);
+        final_url = lr_add_url_protocol(url, err);
+        if (!final_url)
             return FALSE;
-        }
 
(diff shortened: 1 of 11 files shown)


Tojaj:
Hi @sourcejedi 
thank you for your contribution! The patch looks good, but I don't understand what kind of problem it should solve?

What exactly is wrong on working with local repos with relative paths?

    #!/usr/bin/env python
    import librepo
    h = librepo.Handle()
    h.repotype = librepo.YUMREPO
    h.urls = ["../"]
    h.perform()

sourcejedi:
As per commit message and previous PR, my specific concern was that relative paths _in config files_ do not make sense.

You're saying I might have been looking at the wrong level to fix this :). I expect you're right, although I don't have time to really check for a while.

Tojaj:
Understand, in config files, the relative paths really don't make a sense. But if the handle is configured directly by, for example, command arguments, then it may be handy that it it possible to resolve relative paths.
I really appreciate your contribution and time you invest into it, but I don't think it's right to do such change into standard behavior without strong rationale (people may already depend on it and we would be breaking behavior of their apps).

PR #102, opened 2017-03-11
New test: check downloads are failed if SHA512 doesn't match

sourcejedi:
It seems nice to include this as a downloader test,
to guard against logic errors in future changes.

(Reminder: downloader tests require passing `-d` to the C test runner).

---

I wrote this test in a bit of a panic.  Only after running it, did I realize the bug Tojaj found in my PR (https://github.com/rpm-software-management/librepo/pull/89#pullrequestreview-14363156) was not _quite_ as bad as I thought :).  (IOW, my bug did not cause this test to fail).


tests/test_downloader.c
```diff
@@ -290,6 +290,100 @@ START_TEST(test_downloader_three_files_with_error)
 }
 END_TEST
 
+START_TEST(test_downloader_checksum)
+{
+    #define NUM_CHECKSUM_TESTS 2
```
ignatenkobrain: why not to just have last element as `NULL`?

sourcejedi: I'm not used to doing that with structs.  Seems like a good idea, see second commit.

(And apologies for rebasing the first commit, I'm not sure what went wrong but I pushed a version that failed to build due to parameter mismatch).

ignatenkobrain:
@dnf-bot retest this please

@sourcejedi can you rebase this please

PR #105, opened 2017-04-05
[v2] New test: check downloads are failed if SHA512 doesn't match

sourcejedi:
It seems nice to include this as a downloader test,
to guard against logic errors in future changes.

(Reminder: downloader tests require passing `-d` to the C test runner).

---

Rebased version of #102


ignatenkobrain:
@dnf-bot r+

PR #108, opened 2017-06-21
build: support installation in alternate dir

jlebon:
This is the same modification as
https://github.com/rpm-software-management/libdnf/pull/136, but for this
repo. It allows one to build and install in an unprivileged dir.


ignatenkobrain:
@dnf-bot r+

PR #109, opened 2017-06-22
headers: fix function declaration isn't a prototype

jlebon:
The correct way to declare a function that takes no arguments is:

    int foo(void);

If we instead use

    int foo();

it means that `foo` takes an unspecified amount of arguments. GCC/clang
will complain about this declaration style nowadays if
`-Wstrict-prototypes` is turned on.


ignatenkobrain:
@dnf-bot r+

PR #113, opened 2017-08-24
Connection initiation callback for md download (RhBug:1048344)

juliusmilan:
https://bugzilla.redhat.com/show_bug.cgi?id=1048344

Adds starttransfercb callback to _librepo.Handle object, which
signalizes start of a single new transfer. Since internally data may be
downloaded from several mirrors, this provides detailed overview about
start of single transfers together with information about
related mirrors, concretely information about maximal number of mirrors to try
and count of mirrors that failed to connect yet.


librepo/downloader.c
```diff
@@ -1123,6 +1123,19 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
     // Add the transfer to the list of running transfers
     dd->running_transfers = g_slist_append(dd->running_transfers, target);
 
+    // Call LrStartTransferCb to inform upper levels about start of new transfer
+    LrStartTransferCb st_cb = target->target->starttranscb;
+    if (st_cb) {
```
jrohel: Try another approach. If you set callback function then libcurl will call your code just before the connect call.
Hint: `curl_easy_setopt(CURL *handle, CURLOPT_SOCKOPTFUNCTION, sockopt_callback)`

juliusmilan:
updated

librepo/downloader.c
juliusmilan: done

librepo/downloader.h
```diff
@@ -100,6 +100,7 @@ lr_download_single_cb(GSList *targets,
                       gboolean failfast,
                       LrProgressCb cb,
                       LrMirrorFailureCb mfcb,
+                      LrStartTransferCb stcb,
```
j-mracek: It looks like API. And I think this is not good to break it if there still could be a compatible alternative.

librepo/downloadtarget.h
```diff
@@ -197,6 +200,7 @@ lr_downloadtarget_new(LrHandle *handle,
                       void *cbdata,
                       LrEndCb endcb,
                       LrMirrorFailureCb mirrorfailurecb,
+                      LrStartTransferCb starttranscb,
```
j-mracek: Again API

librepo/downloader.h
juliusmilan: How would you achieve a compatible alternative?
The possible solution (maybe) I can think about is to call progresscb with some special values, which upper layers can decode / recognize. But I think it would be an obfuscated solution, which contains a non obvious "secrets". IMO it is cleaner to bump a version.

PR #114, opened 2017-08-29
Add librepo.spec file

jrohel:
Taken from dis-git.
Platform-python stuff removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 build/
 TAGS
 tags
-./librepo.spec
 librepo/version.h
 *.pyc
 doc/c/Doxyfile.in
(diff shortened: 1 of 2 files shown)


ignatenkobrain:
Ah, also remove platform-python stuff from spec file.. It is totally Fedora's downstream part.

jrohel:
OK. I removed platform-python stuff and changelog.

Conan-Kudo:
Why are we doing this?

jrohel:
We want to use the same work flow as for DNF. That means librepo.spec file will be in upstream.
Last changes of DNF needs last version of librepo. We want to track requested version on the time of PR.

ignatenkobrain:
I've pushed it to master, but: @jrohel we don't use librepo.spec from upstream in CI, so whatever you will make here will not affect nightlies / CI...

PR #120, opened 2018-02-08
Re-enable test_download_packages_with_resume_02 test

m-blaha:
This test was disabled by commit 01488e507ca89e9.
The problem was with unfinished download from fake werkzeug / flask
server set up for testing. When client went away after
downloading only 10 bytes, the request remained active and
the server died with a Broken pipe error - hence wasn't able
to serve following tests.
(because of passthrough_errors option of werkzeug run_simple method
set to True by flask.run())

Because this test only makes sure, that partial downloads
after failed resume are not modified or deleted, we can
just fake unfinished download file with random content.


j-mracek:
@rh-atomic-bot r+

PR #123, opened 2018-02-20
Expected size != 0 breaks tests

mhatina:
It shouldn't be used until librepo is refactored properly


ignatenkobrain:
Awesome, merging without any reference to bugs after 1 second after sending PR.

Conan-Kudo:
Not cool. :-1:

PR #124, opened 2018-02-28
Add support to libxml2

r4f4:
`librepo` is the only lib bringing in `expat` as dep in a fedora base minimal image. By adding support to `libxml2`, we can get rid of `expat`


cgwalters:
Is there a use case for retaining `expat` support?  Does anyone shipping librepo want it?

Also FWIW, libostree also [has a metalink parser](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-metalink.c) for example that just uses `GMarkup`, which seems fine for this use case.

ignatenkobrain:
@cgwalters if we don't count that librepo must die and we should use just libsolv's `repo_add_xxx()`, then only reason is that someone probably still want to build it against expat... Since librepo is probably not used without libsolv (which has expat/libxml2 switch), then both ideally should use same xml library.

ignatenkobrain:
ping @dmach

ignatenkobrain:
ping @dmach

ignatenkobrain:
ping @dmach

dmach:
@r4f4 could you remove #ifdef LIBXML2 from the code and rewrite it completely to libxml2?
It's better to stick with one implementation which is widely used rather than have unused code paths which are potential source of problems.

Conan-Kudo:
Echoing from the Mageia/OpenMandriva side, I'd much rather have only libxml2 in the package manager stack. Dropping expat is perfectly fine for me. libsolv is compiled to use libxml2 instead of expat in both distributions, so we'd get the same benefit.

r4f4:
@dmach Sure thing. Working on it.

r4f4:
Let me know what you think of the new patch.

r4f4:
ping @dmach

dmach:
I'll do couple tests and if it passes, I'll merge this.

PR #125, opened 2018-03-13
Reports initial downloads to the screen (RhBug:1483859)

j-mracek:
The first downloads like metalink, and repomd.xml were perform silently.

https://bugzilla.redhat.com/show_bug.cgi?id=1483859


mhatina:
@rh-atomic-bot r+

PR #126, opened 2018-03-22
spec file changes needed for CentOS 7 build

m-blaha:


cgwalters:
Seems reasonable, but is anyone actually syncing this specfile with downstreams?

librepo.spec
```diff
@@ -11,7 +11,7 @@
 
 Name:           librepo
 Version:        1.9.0
-Release:        1%{?dist}
+Release:        2%{?dist}
```
ignatenkobrain: no need for this

PR #127, opened 2018-06-12
Add zchunk support

jdieter:
This patchset adds [zchunk](https://github.com/zchunk/zchunk) support to librepo.  It requires that libsolv also have [zchunk support enabled](https://github.com/openSUSE/libsolv/pull/270), and, for it to be effective, the calling application needs to pass the handle a cache directory to find old zchunk files to delta against (I've submitted [a patch to dnf](https://github.com/rpm-software-management/dnf/pull/1107)).


jdieter:
Do please note, that, unlike the libsolv and dnf patches, these patches make zchunk a mandatory dependency.  If you'd rather it be an optional dependency, please let me know.

Conan-Kudo:
@jdieter I suspect that it might be desired for the RHEL side of the house to be optional, but I don't see a reason why we wouldn't want it on in Fedora and Mageia.

j-mracek:
Without a merge of patch for libsolv it will not work?

jdieter:
That's correct.  It will download the zchunk metadata, and send it to libsolv, but without https://github.com/openSUSE/libsolv/pull/270, libsolv won't be able to read the zchunk metadata.

CMakeLists.txt
```diff
@@ -31,6 +31,12 @@ PKG_SEARCH_MODULE(LIBCRYPTO REQUIRED libcrypto openssl)
 PKG_CHECK_MODULES(LIBXML2 libxml-2.0 REQUIRED)
 FIND_PACKAGE(CURL REQUIRED)
 FIND_PACKAGE(Gpgme REQUIRED)
+FIND_LIBRARY(ZCHUNKLIB NAMES zck)
```
ignatenkobrain: this should be pkg_check_modules

dmach:
zck.h should #include <stdbool.h> as the boolean type is part of public header

There are also couple compilation errors, such as:
/usr/include/zck.h:62:29: note: expected 'zckCtx *' {aka 'struct zckCtx *'} but argument is of type 'int'
librepo/downloader.c:1089:11: error: too few arguments to function 'zck_init_adv_read
librepo/downloader.c:1156:31: error: too few arguments to function 'zck_init_read'
librepo/downloader.c:1208:29: error: too few arguments to function 'zck_get_range_char'
- tested with zchunk-devel-0.9.7-2.fc29

I also created PR to make zchunk optional - it's very important to us so we can turn it off on as needed.

jdieter:
Ok, I've gone ahead and merged your PR and added <stdbool.h> to zck.h in zchunk-0.9.10.  This now compiles fine with zchunk-0.9.10+, but it won't work with repositories generated using the current createrepo_c zchunk patches because of the format changes.

jdieter:
@dmach, If you'd prefer these to be in fewer commits, I'll squash them down for you

Conan-Kudo:
@jdieter Could you please do so?

librepo.spec
```diff
@@ -38,6 +43,9 @@ BuildRequires:  libcurl-devel >= 7.19.0
 BuildRequires:  pkgconfig(libxml-2.0)
 BuildRequires:  pkgconfig(libcrypto)
 BuildRequires:  pkgconfig(openssl)
+%if %{with zchunk}
+BuildRequires:  zchunk-devel
```
Conan-Kudo: Use `pkgconfig(zck) >= 0.9.11` here.

jdieter:
Thanks!

Conan-Kudo:
@dmach, @j-mracek: This is now merged into libsolv, so can we get this reviewed and merged, as well as the corresponding PRs for the other components of the stack (as well as `createrepo_c`)?

PR #130, opened 2018-08-02
Support for md5 checksums fixed.

vimusov:
I found a bug. dnf can not synchronize a repository if it is signed with md5. Here is the error:

`librepo.LibrepoException: (21, 'Unknown checksum type "md5" for /var/lib/mock/[...]/[...]-filelists.xml.gz', 'Unknown type of checksum is needed to verify one or more file(s)')
`

So I made this patch. I hope it will be useful.


m-blaha:
Hi, it looks good to me.

PR #131, opened 2018-08-03
Add support for python-gpg

jrohel:
- pygpme was needed in two unit tests
- librepo can handle both pygpgme and python-gpg now
- compatibility with RHEL8


jrohel:
@ignatenkobrain OK. I made changes.

PR #132, opened 2018-08-03
Enable C unit tests on RHEL

m-blaha:
Python tests are still disabled due to unavailability of python-flask
on RHEL 7.


dmach:
@m-blaha could you rebase this PR? The change is quite straightforward and we should merge it in after re-testing.

m-blaha:
@dmach of course. rebased.

PR #135, opened 2018-09-25
[spec] Synchronize spec with downstream

j-mracek:


librepo.spec
```diff
@@ -7,6 +7,15 @@
 %bcond_without python3
 %bcond_without tests
 %endif
+
+%if 0%{?rhel} && 0%{?rhel} > 7
```
m-blaha: Shouldn't be here also Fedora > 29?
%if 0%{?rhel} > 7 || 0%{?fedora} > 29

PR #141, opened 2019-02-27
Only automatically download zchunk metadata for main metadata

jdieter:
Currently librepo attempts to download any available zchunk repodata if libdnf supports zchunk.  Unfortunately, as can be seen in https://pagure.io/dusty/failed-composes/issue/1529, there are situations where libdnf supports zchunk, but the calling application (in this case, dnf) doesn't.

This patch fixes librepo to only _automatically_ request zchunk repodata if it's `primary`, `filelists` or `other`, since those are automatically decompressed by libsolv, which does support zchunk.  If the calling application supports zchunk, it can manually request the zchunk variant of a record from librepo and then failover itself if that record doesn't exist.

### DIFF
--- a/librepo/yum.c
+++ b/librepo/yum.c
@@ -81,7 +81,10 @@ lr_yum_repo_free(LrYumRepo *repo)
 static char *
 get_type(LrYumRepo *repo, const char *type)
 {
-    if (!repo->use_zchunk)
+    if (!repo->use_zchunk ||
+        (g_strcmp0(type, "primary") != 0 &&
+         g_strcmp0(type, "filelists") != 0 &&
+         g_strcmp0(type, "other") != 0))
         return g_strdup(type);
 
     gchar *chk_type = g_strconcat(type, "_zck", NULL);
@@ -177,6 +180,13 @@ lr_yum_switch_to_zchunk(LrHandle *handle, LrYumRepoMd *repomd)
     if (handle->yumdlist) {
         int x = 0;
         while (handle->yumdlist[x]) {
+            if(g_strcmp0(handle->yumdlist[x], "primary") != 0 &&
+               g_strcmp0(handle->yumdlist[x], "filelists") != 0 &&
+               g_strcmp0(handle->yumdlist[x], "other") != 0) {
+                x++;
+                continue;
+            }
+
             char *check_type = g_strconcat(handle->yumdlist[x], "_zck", NULL);
             assert(check_type);
 


jdieter:
I realize I've given you a number of patches to review recently, but I'd really appreciate it if this one could make it into Rawhide as soon as possible.  It will permanently fix the compose failures that we're hitting there.

AdamWill:
Are you sure this is right? At least as far as I could follow it, the problematic code path was purely between dnf and libdnf, librepo and libsolv were not involved at all. Or will this prevent inclusion of zchunk-encrypted comps in the repodata / repomd at compose time? Thanks!

jdieter:
For comps, the code currently goes something like this:
```
dnf: I want group_gz
    |
    v
libdnf: Let's get group_gz from the repo (BTW, I support zchunk)
    |
    v
librepo: Ooh, we have new shiny group_gz_zck available, why not give you that instead
    |
    v
libdnf: Got zchunked group_gz, thanks
    |
    v
dnf: WTF am I supposed to do with this?
```
With this patch, the code will do:
```
dnf: I want group_gz
    |
    v
libdnf: Let's get group_gz from the repo (BTW, I support zchunk)
    |
    v
librepo: Ok, here's xz'd group_gz, even though you support zchunk
    |
    v
libdnf: Got xz'd group_gz, thanks
    |
    v
dnf: Everything is awesome!
```
When we add support for decompressing zchunked files to dnf, the code will do:
```
dnf: I want group_zck
    |
    v
libdnf: Let's get group_zck from the repo (BTW, I support zchunk)
    |
    v
librepo: Ok, here's zchunked group_zck
    |
    v
libdnf: Got zchunked group_zck, thanks
    |
    v
dnf: Everything is awesome!
```
I've tested that this actually works, and if you'd like to, here are some scratch builds of librepo:
F31: https://koji.fedoraproject.org/koji/taskinfo?taskID=33114133
F30: https://koji.fedoraproject.org/koji/taskinfo?taskID=33114143

If you'd need a test repository with zchunk-compressed comps.xml, you can use https://www.jdieter.net/downloads/repotest/.  To test, you'll need to run some group commands (i.e. `dnf group list`)

Please note that this repository only has the metadata, not the actual RPMs.

jrohel:
@jdieter Thanks for patch. But I think that better will be to add zchunk support into DNF.
So, I started work on it: https://github.com/rpm-software-management/dnf/pull/1341/files

jdieter:
That looks great, and works well around the fact that zchunk doesn't yet have python bindings.  My one concern is about other users of libdnf.  Without this patch, they'll need to also make sure they're using libdnf's decompression functions.

I'm happy to do the work if you can point me to a list of known users.

AdamWill:
From Fedora package repos, at least, the users seem to be only PackageKit, microdnf and hawkey (which is part of libdnf anyhow). dnfdaemon seems not to use libdnf directly but python3-dnf.

Conan-Kudo:
I deliberately try to use `python3-dnf` and avoid the lower layers because it avoids problems like this. There are cases where I poke the lower layers, but that's thankfully not in this case.

jdieter:
Ok, I've tested both PackageKit and microdnf against with all the current packages in Fedora, and both worked fine with zchunked comps and updateinfo.  microdnf only cares about the primary (and possibly filelists), while PackageKit downloads uncompressed comps.xml and works just fine with zchunk-compressed updateinfo.

My one remaining concern is that dnf will work with zchunk-compressed prestodelta.

jdieter:
I've taken a brief look at hawkey, but don't see where it downloads anything, so I'm not too worried about it at the moment.

Conan-Kudo:
@jdieter the hawkey component of libdnf doesn't know how to fetch metadata, as it was purely an abstraction on top of libsolv.

jrohel:
@jdieter The zchunk support was merger into DNF and libdnf. https://github.com/rpm-software-management/dnf/pull/1341, https://github.com/rpm-software-management/libdnf/pull/690.
Thanks for your tests.
I close this PR.

jdieter:
@jrohel, yep, I like your solution much better!  :)  Can we get some builds into F30 and Rawhide so we don't have to keep blocking createrepo_c-0.12.1 during the compose process?

AdamWill:
Note, F30 Beta freeze and Bodhi enablement should've just happened, so if that hasn't been delayed, you'll need to propose a freeze exception to get the F30 build in. See https://fedoraproject.org/wiki/QA:SOP_freeze_exception_bug_process , long story short, we need an RHBZ covering the situation, then either manually set it to block the bug `BetaFreezeException` or use https://qa.fedoraproject.org/blockerbugs/propose_bug .

jdieter:
@AdamWill, thanks for the heads up.  The build didn't make it in, so, if we decide we want this to go in before beta, we'll go ahead and follow the exception process.

My one concern about delaying this is that it increases the number of people who will have the broken combination of librepo and dnf when there's zchunk-compressed comps, but the librepo that tries to read zchunk-compressed comps came out almost three weeks ago, so I'm not sure that it will have any additional effect.

The good news is that, if someone ends up with the broken combination, it's only "dnf group" commands that are broken, and dnf update still works perfectly.

jdieter:
And, of course, most of the zchunk savings comes from primary and filelists, so we can delay createrepo_c-0.12.1, if that helps the situation.

Conan-Kudo:
@jdieter Push comes to shove, I can pull back the patches into the appropriate packages and propose them as exceptions to push through. The process isn't that bad.

jdieter:
@Conan-Kudo, I think that will depend on whether @jrohel would prefer to push out a normal release.

jrohel:
We will push a new release to rawhide next week.

cgwalters:
> From Fedora package repos, at least, the users seem to be only PackageKit, microdnf and hawkey (which is part of libdnf anyhow). dnfdaemon seems not to use libdnf directly but python3-dnf.

FWIW rpm-ostree uses a vendored libdnf copy, and it does use librepo directly.  We are working on rebasing to master here: https://github.com/projectatomic/rpm-ostree/pull/1404

jdieter:
> We will push a new release to rawhide next week.

Great.  I think that works for me.  If we can still get a freeze exception for it, I'd love to get it into F30 before beta comes out (mainly because we'll need to be more careful about when we turn zchunked comps back on if we have people who have downloaded the beta without this fix).

jdieter:
> > From Fedora package repos, at least, the users seem to be only PackageKit, microdnf and hawkey (which is part of libdnf anyhow). dnfdaemon seems not to use libdnf directly but python3-dnf.
> 
> FWIW rpm-ostree uses a vendored libdnf copy, and it does use librepo directly. We are working on rebasing to master here: [projectatomic/rpm-ostree#1404](https://github.com/projectatomic/rpm-ostree/pull/1404)

If it's calling librepo directly and isn't explicitly setting the cache directory (see https://github.com/rpm-software-management/libdnf/pull/478/commits/bb6394e9d3876576ef822ec41dd89f12f2845ca3 for an example), librepo should automatically fall back to non-zchunked metadata.  Your version of libdnf is old enough that you shouldn't need to worry about it trying to pull in zchunked metadata.

Once you've rebased to master, you'll just want to make sure that you use the new function in https://github.com/rpm-software-management/libdnf/pull/690 to decompress any extra files you get from libdnf.  If you want me to take a look when you're closer to finishing it up, just ping me.

jdieter:
> We will push a new release to rawhide next week.

I see this is in updates-testing now. @jrohel, @dmach would you mind if I requested a freeze exception for this build, or are the other changes big enough to wait until after Beta comes out?

PR #144, opened 2019-03-22
Simplified lr_perform()'s loop to prevent busy wait

ydroneaud:
Calling curl_multi_perform() in a loop is actually
busy waiting for data being available on the socket.

The whole logic in lr_perform() can be simplified
allowing more efficient processing. The function has
to repeatedly do:

1) call curl_multi_perform() for curl to fetch data if needed/available
2) call check_transfer_statuses() to check download status
3) if curl needs more data, wait for the socket to have some

Once there's no need for more data, and there's no other
transfer pending, lr_perform() can return TRUE.

The function can be further simplified by
- assuming curl_multi_perform() is not returning CURLM_CALL_MULTI_PERFORM
  require cURL 7.20, https://curl.haxx.se/libcurl/c/curl_multi_perform.html
- using curl_multi_wait() instead of select()
  require cURL 7.28, https://curl.haxx.se/libcurl/c/curl_multi_wait.html

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1691856

Fixes: 208edbecf3b0 ('Reduce download delays by using still_running correctly.')

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -2418,43 +2418,41 @@ static gboolean
 lr_perform(LrDownload *dd, GError **err)
 {
     CURLMcode cm_rc;    // CurlM_ReturnCode
-    int still_running;
 
     assert(dd);
     assert(!err || *err == NULL);
 
-    do { // Before version 7.20.0 CURLM_CALL_MULTI_PERFORM can appear
+    while (1) {
+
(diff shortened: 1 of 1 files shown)


jrohel:
Thank you. Fix with nice simplification.

PR #147, opened 2019-03-27
[tests] Wait for server to start instead of doing arbitrary sleep

pkratoch:
The librepo build occasionally failed on TestCaseYumPackageDownloading with
error "Curl error (7): Couldn't connect to server..."

### DIFF
--- a/tests/python/tests/base.py
+++ b/tests/python/tests/base.py
@@ -3,6 +3,7 @@ import time
 import socket
 import ctypes
 import os.path
+import requests
 from multiprocessing import Process, Value
 from tests.servermock.server import app
 try:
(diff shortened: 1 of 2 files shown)


tests/python/tests/base.py
```diff
@@ -93,9 +94,19 @@ class TestCaseWithFlask(TestCase):
...
         cls.server = Process(target=application, args=(cls._TS_PORT,))
         cls.server.start()
-        time.sleep(0.5)
         cls.MOCKURL = MOCKURL_TEMPLATE % cls._TS_PORT.value
         cls.PORT = cls._TS_PORT.value
+        # Wait for the server to start (max 5 seconds)
+        for i in range(50):
+            time.sleep(0.1)
+            try:
+                requests.get(cls.MOCKURL)
```
m-blaha: There should be a timeout set for requests.get() to prevent hanging:
http://docs.python-requests.org/en/master/user/quickstart/#timeouts

pkratoch: Thanks, I added that.

pkratoch:
@lukash, ok, you are right about the timeout, so I used your solution (I can't see your comment anymore). Thanks for review.

lukash:
One day I'll find out why some comments on outdated commits stay and some disappear :smile: 

Thanks for the patch, merging.

PR #148, opened 2019-03-31
Clean up target->curl_handle rather than target->handle

jdieter:
Cleaning up target->handle causes a segfault when this codepath is run, so this patch fixes it so we clean up target->curl_handle.

The segfault is reported at https://bugzilla.redhat.com/show_bug.cgi?id=1694411 and affects all updates where zchunked metadata is available.


jrohel:
Thanks for fix.

PR #149, opened 2019-05-02
Add an option to preserve timestamps of the downloaded files (RhBug:1688537)

m-blaha:
If LRO_PRESERVETIME is set to True, librepo will try to keep the timestamps
of downloaded files in sync with that on the remote side.

https://bugzilla.redhat.com/show_bug.cgi?id=1688537


librepo/downloader.c
```diff
@@ -18,7 +18,8 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
-#define _XOPEN_SOURCE   500 // Because of fdopen() and ftruncate()
+#define _BSD_SOURCE
+#define _DEFAULT_SOURCE
```
Conan-Kudo: Why did you drop `_XOPEN_SOURCE 500`?

m-blaha: I need futimes() to change file timestamps and this requires _DEFAULT_SOURCE (for glibc > 2.19) or _BSD_SOURCE (for glibc <= 2.19, which is the case for RHEL7).
Having _XOPEN_SOURCE explicitly defined means, that _DEFAULT_SOURCE is not defined.

Conan-Kudo: How does this affect non-glibc environments?

m-blaha: I'm not sure. But I can return _XOPEN_SOURCE   500 back

Conan-Kudo: It shouldn't hurt anything to put it back, and I'm wary of accidentally breaking non-glibc builds (there are a few out there, including Yocto...).

Conan-Kudo:
@rh-atomic-bot r+

jamescassell:
Seems like it would be better to enable this option by default or always use this Behavior with an out an option to closer match python URL grabber from yum 3

PR #150, opened 2019-05-03
librepo: append the '?' part of repo URL after the path

alesax:
For certain repository systems, like SCC, URL contains '?' part with
some identification token. As librepo constructs the URL by mechanically
appending file path after it, it does not work.

This naive patch takes the '?' part and appends it at the end of URL.

i.e. instead of

  https://updates.suse.com/SLE15?SeCrEtToKeN/repodata/repomd.xml

there's something like:

  https://updates.suse.com/SLE15/repodata/repomd.xml?SeCrEtToKeN

This enables librepo users (like dnf) to work with SCC repositories.

Signed-off-by: Ales Novak <alnovak@suse.cz>

### DIFF
--- a/librepo/util.c
+++ b/librepo/util.c
@@ -187,6 +187,7 @@ lr_pathconcat(const char *first, ...)
     size_t total_len;  // Maximal len of result
     size_t offset = 0;
     int is_first = 1;
+    char *qmark_section;
     int previous_was_empty = 0; // If last chunk was "" then separator will be
                                 // appended to the result
 
@@ -202,6 +203,8 @@ lr_pathconcat(const char *first, ...)
 
     if (total_len == 0)
         return g_strdup("");
+
+    qmark_section = strchr(first, '?');
 
     res = lr_malloc(total_len + separator_len + 1);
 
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@alesax Merged, thanks for the patch!

lukash:
Hello, the function is used to operate on filesystem paths as well. Question mark is a valid character without a special meaning in this context. However unlikely, this patch is going to break operations on FS paths containing a question mark. This needs to be fixed.

As a separate point, adding this functionality changes the behaviour of the function away from a simple concatenation, so the name should be changed.

It would also be nice to add some documentation of the new and somewhat unexpected behaviour, so that a reader doesn't have to figure it out from the code, as it's not trivially apparent.

@Conan-Kudo

alesax:
@lukash you're right about local fs, would you agree with fix along these lines (appearing in different place of code)?

```
-    qmark_section = strchr(first, '?');
+    qmark_section = lr_is_local_path(first)?NULL:strchr(first, '?');
```

lukash:
@alesax I'm not fond of it, its getting kind of funky.

I would much prefer to be explicit about it and create another function called along the lines of `lr_url_add_path`, which would add the question mark magic and would be used for URLs only. The `lr_pathconcat` function can be kept for concating local paths and unless I'm mistaken can be used in the new `lr_url_add_path`, you just need to cut off the arguments part and append it again later.

This way it will be more clear to see there's something going on with the `?` and what it is.

If you're going to change it, can you please add (a possibly condensed version of) the description from the commit message into a (doc) comment to the function?

Thanks for coming back to this.

PR #152, opened 2019-05-13
Don't create socket during repo key import (RhBug:1650266)

jrohel:
It configures the gpg agent to run in server mode and to wait for stdin commands. The default gpg agent mode is to create a socket and listen for commands there.


j-mracek:
@rh-atomic-bot r+

PR #154, opened 2019-05-21
Exit gpg-agent after repokey import (RhBug:1650266)

jrohel:
It tries to exit gpg-agent. Path to the communication socket is derived from homedir. The gpg-agent automaticaly removes all its socket before exit.
Newer gpg-agent creates sockets under [/var]/run/user/{pid}/... if directory exists. In this case gpg-agent will not be exited.
It solves the https://bugzilla.redhat.com/show_bug.cgi?id=1650266 issue of remaining sockets in the gpg home directory.

This PR replaces another solution https://github.com/rpm-software-management/librepo/pull/152 that configures the gpg agent to run in server mode (without sockets). Importing of repo keys was very slow  with gpg-agent in server mode during the unit tests.

### DIFF
--- a/librepo/gpg.c
+++ b/librepo/gpg.c
@@ -32,6 +32,30 @@
 #include "util.h"
 #include "gpg.h"
 
+static void
+kill_gpg_agent(gpgme_ctx_t context, const char *home_dir)
+{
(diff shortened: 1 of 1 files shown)


j-mracek:
LGTM!

PR #155, opened 2019-06-14
Handle webservers that don't support ranges when downloading zck

jdieter:
Make sure we fall back to downloading full zchunk file if a webserver doesn't support ranges.  Current code just fails to next mirror.


lukash:
Tested locally. Thanks Jonathan.

PR #158, opened 2019-07-07
Retry downloads when baseurl or full path provided

j-mracek:


librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
-                    __func__, mirrorurl);
-            continue;
-        }
+            if (target->handle
+                && target->handle->offline
+                && c_mirror->mirror->protocol != LR_PROTOCOL_FILE)
+            {
+                if (!maximal_errors) {
+                // Skip each url that doesn't have "file://" or "file:" prefix
+                g_debug("%s: Skipping mirror %s - Offline mode enabled", __func__, mirrorurl);
```
lukash: Missing indent.

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
+            *selected_mirror = c_mirror;
+            return TRUE;
         }
-
-        // This mirror looks suitable - use it
-        *selected_mirror = c_mirror;
-        return TRUE;
-    }
+    } while (g_slist_length(target->tried_mirrors) < dd->allowed_mirror_failures &&
+    ++maximal_errors < dd->allowed_mirror_failures);
```
lukash: Might be personal preference, but why not use a `while` instead of `do ... while`, makes it easier to read and there's no benefit to `do` here I think.

Also, the whole mirror-picking mechanism seems to be complex already and adding one more loop around it pushes it to being quite convoluted.

Why not make it simply loop over the available mirrors until either success or number of retries is reached?

```C++
GSList *elem = target->lrmirrors;
int num_retires = 0;
while (elem != NULL && num_retires <= dd->allowed_mirror_failures) {
    if (try_mirror(...)) {
        break;
    }

    num_retries++;
    elem = g_slist_next(elem);
    // loop around in case we reach the end of the mirror list
    if (elem == NULL) {
        elem = target->lrmirrors;
    }
}
```

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
+            LrMirror *c_mirror = elem->data;
+            gchar *mirrorurl = c_mirror->mirror->url; // shortcut
+            
+            if (!maximal_errors) {
+                if (!maximal_errors && g_slist_find(target->tried_mirrors, c_mirror)) {
+                    // This mirror was already tried for this target
+                    continue;
+                }
+                if (c_mirror->successful_transfers == 0 &&
+                    dd->allowed_mirror_failures > maximal_errors &&
```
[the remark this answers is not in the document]
j-mracek: I believe that this is correct or there is some misunderstanding.

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
-
-        if (c_mirror->mirror->protocol == LR_PROTOCOL_RSYNC) {
-            // Skip rsync mirrors
-            g_debug("%s: Skipping rsync url: %s", __func__, mirrorurl);
-            continue;
-        }
+            if (c_mirror->mirror->protocol == LR_PROTOCOL_RSYNC) {
+                if (!maximal_errors) {
+                    // Skip rsync mirrors
+                    g_debug("%s: Skipping rsync url: %s", __func__, mirrorurl);
```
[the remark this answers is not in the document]
j-mracek: I only log it during the first iteration.

librepo/downloader.c
j-mracek: I am sorry but I have a feeling that the proposed approach will not deliver the required behaviour.

librepo/downloader.c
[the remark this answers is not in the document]
lukash: You're inside a condition here in which you've checked that `maximal_errors` is equal to 0.

librepo/downloader.c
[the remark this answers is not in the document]
lukash: Oh, right. I got it mixed up, sorry.

librepo/downloader.c
lukash: What _is_ the required behaviour though? I have a feeling the current code is unnecessarily complex.

j-mracek: @lukash From my point of view do {} while () loop is more readable that proposed change. The behaviour should allow to use all mirrors first. Then to allow of reuse of limited number of mirrors.
The construct directly says it. Please could you specify whatever it is no go request?

lukash: > @lukash From my point of view do {} while () loop is more readable that proposed change.

Readability is subjective, but I don't think you can really say that your code is more readable than what I proposed. You have two nested loops there with several more nested `if`s checking multiple conditions. Granted my example is simplified and will surely need more stuff added, it is a simple logical flow of what you'd expect the function to do. So I find it hard to believe you think your code is more readable...

> The behaviour should allow to use all mirrors first. Then to allow of reuse of limited number of mirrors.

This is too vague and in fact my example can do exactly the same (by adding a single `if`). I believe that if you want to merge code like this, you should be able to describe with words what it is supposed to be doing and actually put that into a comment. Specifically focusing on the reasons why it actually needs to be complex. One thing is making it easier to understand for others, the other thing is, by adding complexity, you are adding more corner cases and more possibilities for bugs. (And we should really be avoiding this, that's why I'm having this argument)

Actually, you have two nested loops there, while there is no code in the outer loop (apart from incrementing `maximal_errors`) and both loops iterate over the same list of items. So I think it can be proven theoretically you can achive the same with a single loop :slightly_smiling_face:.

> The construct directly says it. Please could you specify whatever it is no go request?

I'm not sure I understand what you mean here, but if you're asking whether I insist on changing this in the PR, I think so, until I understand the reasons it needs to be like this...

Cheers!

librepo/downloader.c
```diff
@@ -2331,11 +2342,15 @@ check_transfer_statuses(LrDownload *dd, GError **err)
                     target->tried_mirrors = g_slist_remove(target->tried_mirrors, target->mirror);
                     num_of_tried_mirrors = g_slist_length(target->tried_mirrors);
                 }
-
-                if (can_try_more_mirrors(dd, num_of_tried_mirrors))
+                gboolean complete_url_or_baseurl = complete_url_in_path || target->target->baseurl;
```
praiskup: everytime when I see `baseurl` here this function, the code is entirely misleading -- because baseurl isn't only about `baseurl=` statement, but it can mean link to mirrorlist or metalink.

librepo/downloader.c
```diff
@@ -691,64 +694,74 @@ select_suitable_mirror(LrDownload *dd,
...
 
     *selected_mirror = NULL;
+    // mirrors_iterated is used to allow to use mirrors multiple times for a targed
+    int mirrors_iterated = 0;
+    do {
+        // Iterate over mirror for the target
+        for (GSList *elem = target->lrmirrors; elem; elem = g_slist_next(elem)) {
+            LrMirror *c_mirror = elem->data;
+            gchar *mirrorurl = c_mirror->mirror->url; // shortcut
+
```
praiskup: here is trailing whitespace

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
     assert(!err || *err == NULL);
 
     *selected_mirror = NULL;
+    // maximal errors is used to allow use mirrorst multiple times for a targed
```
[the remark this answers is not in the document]
j-mracek: Resolved

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
+    int maximal_errors = 0;
+    do {
+        // Iterate over mirror for the target
+        for (GSList *elem = target->lrmirrors; elem; elem = g_slist_next(elem)) {
+            LrMirror *c_mirror = elem->data;
+            gchar *mirrorurl = c_mirror->mirror->url; // shortcut
+            
+            if (!maximal_errors) {
+                if (!maximal_errors && g_slist_find(target->tried_mirrors, c_mirror)) {
```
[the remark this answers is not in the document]
j-mracek: Resolved

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
...
     *selected_mirror = NULL;
+    // maximal errors is used to allow use mirrorst multiple times for a targed
+    int maximal_errors = 0;
+    do {
+        // Iterate over mirror for the target
+        for (GSList *elem = target->lrmirrors; elem; elem = g_slist_next(elem)) {
+            LrMirror *c_mirror = elem->data;
+            gchar *mirrorurl = c_mirror->mirror->url; // shortcut
+            
+            if (!maximal_errors) {
```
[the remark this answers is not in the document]
j-mracek: Resolved

librepo/downloader.c
```diff
@@ -691,64 +691,74 @@ select_suitable_mirror(LrDownload *dd,
     assert(!err || *err == NULL);
 
     *selected_mirror = NULL;
+    // maximal errors is used to allow use mirrorst multiple times for a targed
+    int maximal_errors = 0;
```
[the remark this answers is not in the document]
j-mracek: resolved, the value was renamed to mirrors_iterated

librepo/downloader.c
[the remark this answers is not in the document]
j-mracek: Resolved

librepo/downloader.c
j-mracek: Resolved

librepo/downloader.c
j-mracek: Yes I know. Even I believe that baseurl could be completely removed from the code, because it is handled as a single mirror. But it means to rewrite a lot of code.

librepo/downloader.c
j-mracek: Fixed

librepo/downloader.c
j-mracek: I added comment to provide additional hint why the loop is there. I don't have a good feeling to change the loop system here. Especially in case that the change request is marked as "Might be personal preference".
But I have a different question: Shall we add some wait step after the first iteration?

librepo/downloader.c
```diff
@@ -691,64 +691,77 @@ select_suitable_mirror(LrDownload *dd,
...
+                if (c_mirror->successful_transfers == 0 &&
+                    dd->allowed_mirror_failures > 0 &&
+                    c_mirror->failed_transfers >= dd->allowed_mirror_failures)
+                {
+                    // Skip bad mirrors
+                    g_debug("%s: Skipping bad mirror (%d failures and no success): %s",
+                            __func__, c_mirror->failed_transfers, mirrorurl);
+                    continue;
+                }
+            } else if (mirrors_iterated < c_mirror->failed_transfers) {
```
lukash: So IIUC, mirrors can pick up `failed_transfers` on different targets that were downloaded previously and can thus always end up being skipped for the subsequent downloads.

Example:
I have one mirror and two targets. The first target is not on the mirror (404), the second is there. The first target will be attempted 4 times and raise `c_mirror->failed_transfers` to 4. For the second target the mirror will never be tried, because the conditions in the code above will always rule it out. Is this correct?

librepo/downloader.c
```diff
@@ -2344,11 +2345,15 @@ check_transfer_statuses(LrDownload *dd, GError **err)
                     target->tried_mirrors = g_slist_remove(target->tried_mirrors, target->mirror);
                     num_of_tried_mirrors = g_slist_length(target->tried_mirrors);
                 }
-
-                if (can_try_more_mirrors(dd, num_of_tried_mirrors))
+                gboolean complete_url_or_baseurl = complete_url_in_path || target->target->baseurl;
+                if (can_try_more(dd, num_of_tried_mirrors, complete_url_or_baseurl))
```
lukash: It feels like this function does two independent things now and is no longer very helpful. Consider putting the contents of the function directly in place here? I don't insist on this one though.

librepo/downloader.c
lukash: It is not just a "Might be personal preference", so please don't dismiss it as that.

I've fixed up your comment and added a couple more while reviewing the code, please have a look and if you don't mind, squash it with your commit: https://github.com/lukash/librepo/commit/8d4dd9f8618fa64b37d17c0e65e9180294fa9d71

As for the wait, if you mean to add it to this loop you've added to select_suitable_mirror(), then I'd say certainly not, as that function only selects the mirror for the next download, so IMHO you'd just be messing up the flow by adding random waits. But I'm still barely understanding the code, so...

librepo/downloader.c
j-mracek: Yes, it is correct.

lukash: Well, then it seems like a problem with the algorithm, as it does not try every mirror at least once for each target, which I think it should?

librepo/downloader.c
```diff
@@ -293,8 +293,11 @@ is_max_mirrors_unlimited(const LrDownload *download)
 }
 
 static gboolean
-can_try_more_mirrors(const LrDownload *download, int num_of_tried_mirrors)
+can_try_more(const LrDownload *download, int num_of_tried_mirrors, gboolean complete_tath_or_baseurl)
```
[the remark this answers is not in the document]
j-mracek: Corrected

librepo/downloader.c
j-mracek: The algorithm is correct, because there are two contradicting requirements:
  a. try to download everything
  b. fail fast

librepo/downloader.c
j-mracek: The function was renamed and added a description for the function. Hope that it helps.

librepo/downloader.c
```diff
@@ -293,7 +293,15 @@ is_max_mirrors_unlimited(const LrDownload *download)
 }
 
 static gboolean
-can_try_more(const LrDownload *download, int num_of_tried_mirrors, gboolean complete_path_or_baseurl)
+/**
+ * @brief Can be tried another mirror or retried baseurl or fullpath
+ *
+ * @param download p_download:...
```
lukash: Is this some special syntax with the `p_download:...`? What does it do?

librepo/downloader.c
```diff
@@ -293,7 +293,15 @@ is_max_mirrors_unlimited(const LrDownload *download)
...
+ * @param num_of_tried_mirrors p_num_of_tried_mirrors:...
+ * @param complete_path_or_baseurl p_complete_path_or_baseurl: determine type of download - mirrors or baseurl/fullpath
+ * @return gboolean Return TRUE when another chance to download is allowed.
+ */
+can_try_more_paths(const LrDownload *download, int num_of_tried_mirrors, gboolean complete_path_or_baseurl)
```
lukash: I don't think the `_path` suffix si good, how about naming the function `can_retry_download`?

librepo/downloader.c
```diff
@@ -293,7 +293,15 @@ is_max_mirrors_unlimited(const LrDownload *download)
 }
 
 static gboolean
-can_try_more(const LrDownload *download, int num_of_tried_mirrors, gboolean complete_path_or_baseurl)
+/**
+ * @brief Can be tried another mirror or retried baseurl or fullpath
```
lukash: This description isn't great, how about:
> Returns whether the download can be retried, using the same URL in case of baseurl or full path, or using another mirror in case of using mirrors.

librepo/downloader.c
j-mracek: not sure, this was provided directly by kdevelop

librepo/downloader.c
```diff
@@ -2349,8 +2357,11 @@ check_transfer_statuses(LrDownload *dd, GError **err)
                     target->tried_mirrors = g_slist_remove(target->tried_mirrors, target->mirror);
                     num_of_tried_mirrors = g_slist_length(target->tried_mirrors);
                 }
+                // complete_url_in_path and target->baseurl doesn't have an alternatives like using
+                // mirrors, therefore they are handled differently and conditions for retry download
+                // also differes
```
lukash: At the end, i'd phrase it:
> ... conditions for retry also differ

(the fact that it's a download retry is clear from the context)

librepo/downloader.c
lukash: Oh. Then it's better to remove it? The ... probably means you should add your own description? and I'm not sure what the `p_*` but I don't think we're using it, so let's not add it, it's redundant and useless?

FWIW I find the argument names self-describing, reiterating what the name already says in the argument description has little value. Since this is not a public function, I'd say a simple brief description is sufficient?

lukash:
Merging, thanks :slightly_smiling_face:

praiskup:
Thanks guys, any plans for new release?

pkratoch:
@praiskup, I will release it to Fedora rawhide, 30 and 29 probably today or tomorrow. I will post the bodhi update links here when it's done.

pkratoch:
@praiskup, sorry for a delay.

After merging this PR, librepo build fails on test test_download_packages_with_resume_02. After a discussion with j-mracek, we decided to skip this test teporarily, since the issue is not severe and the fix will take some time, while this PR is more urgent. I made a PR https://github.com/rpm-software-management/librepo/pull/160 to skip the test.

pkratoch:
Fedora 30: https://bodhi.fedoraproject.org/updates/FEDORA-2019-13479b151a
Fedora 29: https://bodhi.fedoraproject.org/updates/FEDORA-2019-b11742ac7d

praiskup:
Today we observed this:
```
2019-08-16T12:04:48Z DEBUG error: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64 (https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64).
2019-08-16T12:04:48Z DEBUG error: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64 (https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64).
2019-08-16T12:04:48Z DEBUG error: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64 (https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64).
2019-08-16T12:04:48Z DEBUG error: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64 (https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64).
2019-08-16T12:04:48Z DEBUG Cannot download 'https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64': Cannot prepare internal mirrorlist: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-30&arch=aarch64.
2019-08-16T12:04:48Z ERROR Failed to download metadata for repo 'fedora'
2019-08-16T12:04:48Z DDEBUG Cleaning up.
```
The metalink download should poll around the CNAMEs ... though, I'm curious whether we shouldn't put there some `sleep` call, according to the timestamp all the requests were done at the same time.

praiskup:
Does librepo/libcurl name resolution for each attempt, or are there some caches?

PR #159, opened 2019-07-15
retry mirrorlist/metalink downloads

praiskup:
Previously librepo did only one attempt to download the mirror
files.  And because especially Fedora metalink urls are nowadays
by default downloaded from 'mirrors.fedoraproject.org' host -
which is just a CNAME to set of DNS-load balanced proxies - it not
unlikely that librepo would ask some temporarily broken proxy.
Several attempts though will usually mean that librepo will
approach several different proxies.

In manual dnf-workflow, user would just re-try the whole dnf
command manually, but in build systems like Copr [1] this causes
severe problems and random build failures.

[1] https://pagure.io/fedora-infrastructure/issue/7987

### DIFF
--- a/librepo/handle.c
+++ b/librepo/handle.c
@@ -789,6 +789,28 @@ lr_handle_prepare_urls(LrHandle *handle, GError **err)
 }
 
 static gboolean
+lr_yum_download_url_retry(int attempts, LrHandle *lr_handle, const char *url,
+                          int fd, gboolean no_cache, gboolean is_zchunk,
+                          GError **err)
+{
+    gboolean ret = FALSE;
+
+    for (int i = 1;; i++) {
+        ret = lr_yum_download_url(lr_handle, url, fd, no_cache, is_zchunk, err);
+        if (ret)
+            return ret;
+
+        if (i >= attempts)
+            return ret; // Caller to handle the last err
+
+        g_debug("%s: Attempt #%d to download %s failed: %s",
+                __func__, i, url, (*err)->message);
+        ftruncate(fd, 0);
+        g_clear_error (err);
+    }
+}
+
+static gboolean
 lr_handle_prepare_mirrorlist(LrHandle *handle, gchar *localpath, GError **err)
 {
     assert(handle->mirrorlist_fd == -1);
@@ -845,7 +867,7 @@ lr_handle_prepare_mirrorlist(LrHandle *handle, gchar *localpath, GError **err)
         }
 
         url = lr_prepend_url_protocol(handle->mirrorlisturl);
-        if (!lr_yum_download_url(handle, url, fd, TRUE, FALSE, err)) {
+        if (!lr_yum_download_url_retry(3, handle, url, fd, TRUE, FALSE, err)) {
             close(fd);
             return FALSE;
         }
@@ -960,7 +982,7 @@ lr_handle_prepare_metalink(LrHandle *handle, gchar *localpath, GError **err)
         }
 
         url = lr_prepend_url_protocol(handle->metalinkurl);
-        if (!lr_yum_download_url(handle, url, fd, TRUE, FALSE, err)) {
+        if (!lr_yum_download_url_retry(3, handle, url, fd, TRUE, FALSE, err)) {
             close(fd);
             return FALSE;
         }


librepo/handle.c
```diff
@@ -788,6 +788,29 @@ lr_handle_prepare_urls(LrHandle *handle, GError **err)
     return TRUE;
 }
 
+static gboolean
+lr_download_url_retry(LrHandle *lr_handle, const char *url, int fd,
```
dmach: I'd prefer to preserve all arguments from the original function, ideally in the same order:
```
lr_yum_download_url(LrHandle *lr_handle, const char *url, int fd,
                    gboolean no_cache, gboolean is_zchunk, GError **err);
```
The new function could look like:
```
lr_yum_download_url_retry(int attempts, LrHandle *lr_handle, const char *url, int fd,
                    gboolean no_cache, gboolean is_zchunk, GError **err);
```

librepo/handle.c
```diff
@@ -788,6 +788,29 @@ lr_handle_prepare_urls(LrHandle *handle, GError **err)
...
+        if (i >= attempts)
+            return ret; // Caller to handle the last err
+
+        g_debug("%s: Attempt #%d to download %s failed: %s",
+                __func__, i, url, (*err)->message);
+        ftruncate(fd, 0);
+        g_clear_error (err);
+    }
+
+    return ret;
```
praiskup: dead-code return

librepo/handle.c
praiskup: do you want to have the `_yum_` string in function name?

dmach: Yes, keep it there for consistency with the old function name.

lukash:
This seems to be actually solving the same (or similar) issue as https://github.com/rpm-software-management/librepo/pull/159, at least from what I can tell from the description. Although this PR is much more straight-forward.

Anyway, this one hardcodes 3 for the number of attempts... We should probably wait for @j-mracek to come back from vacation (next monday).

praiskup:
#158 seems to deal with iterating over list of mirrors, while this is about getting the list of mirrors, but yeah - similar topic, thanks for noting (I missed that)

praiskup:
Example of affected build from today:
https://copr.fedorainfracloud.org/coprs/g/copr/copr/build/973054/

Snippet from the librepo's log:
```
2019-07-17T07:03:58Z DEBUG lr_yum_use_local_load_base: open(/var/lib/mock/973054-fedora-29-i386-1563347036.746687/root/var/cache/dnf/fedora-b80ee773d578c9a3/repodata/repomd.xml): No such file or directory
2019-07-17T07:03:58Z DEBUG lr_handle_perform: Restoring an old SIGINT handler
2019-07-17T07:03:58Z DEBUG lr_handle_perform: Using dir: /var/lib/mock/973054-fedora-29-i386-1563347036.746687/root/var/cache/dnf/fedora-b80ee773d578c9a3/tmpdir.Bg1l4d
2019-07-17T07:03:58Z DEBUG lr_handle_perform: Using own SIGINT handler
2019-07-17T07:03:58Z DEBUG lr_handle_prepare_internal_mirrorlist: Preparing internal mirrorlist
2019-07-17T07:03:58Z DEBUG lr_download: Target: https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386 (-)
2019-07-17T07:03:58Z DEBUG select_next_target: Selecting mirror for: https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386
2019-07-17T07:03:58Z DEBUG prepare_next_transfer: URL: https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386
2019-07-17T07:03:58Z DEBUG lr_download: Downloading started
2019-07-17T07:03:59Z DEBUG check_transfer_statuses: Transfer finished: https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386 (Effective url: https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386)
2019-07-17T07:03:59Z DEBUG check_transfer_statuses: Error during transfer: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-29&arch=i386
2019-07-17T07:03:59Z DEBUG check_transfer_statuses: No more retries (tried: 1)
```

j-mracek:
@praiskup #158 also resolves your issue. It covers all librepo handler baseurl, fullpath, and mirrors -> fetching mirrors use fullpath handler setting (see commit https://github.com/rpm-software-management/librepo/pull/158/commits/f966ff9efbfd6101ceae03e42fe3f46b77749eb4). My change is much deeper in the code therefore more general and applicable for all download cases (I hope).

praiskup:
I don't think so, these pull-requests are really solving orthogonal things.  We don't have problem with downloading from mirrors, but with downloading metalinks and mirrorlists (the generated files which define what mirrors librepo should use later -- and after that #159 applies).

j-mracek:
@praiskup My approach was to modify `check_transfer_statuses` that is called like:
`lr_yum_download_url -> lr_download_target -> lr_download -> lr_perform -> check_transfer_statuses`

Your was to call `lr_yum_download_url` up to three times.

I also think that #158 cannot be merged with this pull-request, because then mirror will be fetched up to 12x.

praiskup:
Meh, what a reading :-( agreed that #158 is related.  Please merge soon then.

dmnks:
@jrohel So it seems this PR got merged in the end. As a result, we now do 12 retries instead of the original 4 (or 3 as proposed in this PR). Is that intentional?

PR #161, opened 2019-07-26
Adjust test for a new behavior of retrying mirrors

j-mracek:
The test tested that after failed download the resumed file exist with
original content. The original code removed unfinished file before a
second attempt to download removed unfinished file. The new
implementation tries mirrosrs up to allowedmirrorfailures, therefore it
resulted in removal of file before the second try.


pkratoch:
Thanks Jarda, it's definitely better than to skip the test :)
The tests are passing locally with this patch, merging.

PR #162, opened 2019-07-29
Dowload package over damaged rpm

j-mracek:


lukash:
Hi, it's not really clear from the first commit ("Adjust test for a new behavior of retrying mirrors") description why the `h.allowedmirrorfailures = 1` is needed and (without digging into librepo code) I don't see why it should be? (I know the commit is already merged in a different PR).

I'm also not entirely sure this fixes the issue from https://bugzilla.redhat.com/show_bug.cgi?id=1690894 (which is that you write garbage to a partially downloaded file, so a resume will never end up with the right content). Again, didn't go looking right now. But I think that case is not hard to cover with a test, how about adding one?

j-mracek:
> Hi, it's not really clear from the first commit ("Adjust test for a new behavior of retrying mirrors") description why the `h.allowedmirrorfailures = 1` is needed and (without digging into librepo code) I don't see why it should be? (I know the commit is already merged in a different PR).
> 
> I'm also not entirely sure this fixes the issue from https://bugzilla.redhat.com/show_bug.cgi?id=1690894 (which is that you write garbage to a partially downloaded file, so a resume will never end up with the right content). Again, didn't go looking right now. But I think that case is not hard to cover with a test, how about adding one?

The commit was part of another PR. The commit was removed from PR

j-mracek:
With test we have a problem. It cannot be tested in /tmp or our ci-dnf-stack because file system there does not allow to use extended file attribute.

lukash:
> The commit was part of another PR. The commit was removed from PR

Yes, but you didn't answer my question...

> With test we have a problem. It cannot be tested in /tmp or our ci-dnf-stack because file system there does not allow to use extended file attribute.

Yeah... does it not do any resuming if the attribute isn't there?

The test data could be moved to the current directory, which works as long as that doesn't end up being a `tmpfs` (like when the whole build happens in `tmpfs` which AFAIK is what e.g. `tito` does).

j-mracek:
> > The commit was part of another PR. The commit was removed from PR
> 
> Yes, but you didn't answer my question...
> 
> > With test we have a problem. It cannot be tested in /tmp or our ci-dnf-stack because file system there does not allow to use extended file attribute.
> 
> Yeah... does it not do any resuming if the attribute isn't there?

Correct when attribute is not present, librepo always download file from scratch 

> The test data could be moved to the current directory, which works as long as that doesn't end up being a `tmpfs` (like when the whole build happens in `tmpfs` which AFAIK is what e.g. `tito` does).

> Even if test data are moved into current directory it doesn't provide an insurance that the location allows to use extended attributes.

librepo/downloader.c
```diff
@@ -991,9 +991,20 @@ has_librepo_xattr(int fd)
...
 static void
-remove_librepo_xattr(int fd)
+remove_librepo_xattr(LrDownloadTarget * target)
 {
+    int fd = target->fd;
+    if (fd != -1) {
+        fremovexattr(fd, XATTR_LIBREPO);
+        return;
+    }
+    // File descriptor was not provided to targed therefore open file is required
```
lukash: The comment could use some English improvements, but I don't find it particularly helpful - what about something like:
> If file descriptor wasn't set, file name was, and we need to open it

lukash:
I think this can be merged, however, it is most likely unrelated to the bug and won't fix it?

j-mracek:
Thanks for review.

j-mracek:
@rh-atomic-bot r+

j-mracek:
I believe that there is no bot.

PR #164, opened 2019-08-29
logging: raise logging level of error messages

lukash:
Raises logging level of error messages to WARNING for serious local
(mostly filesystem related) errors and to INFO for HTTP errors (for
which it is normal to happen occasionally and are out of the user's
control).

Also logs a single INFO message that looks like:
INFO Downloading: <URL>

Which is logged each time a new file is being downloaded.

For the messages this commit touches I've also removed logging of the
function name (__func__) at the start of the messages, as it has very
little information value even for developers and it throws off the
alignment of the actual logged messages, making the log harder to read.

https://bugzilla.redhat.com/show_bug.cgi?id=1737709

I've tried my best to pick the right error messages (I've left some at debug when it seemed to me the error is handled through GError), but the error handling seems to be in a very bad shape and will need a lot of time and significant changes to untangle - which will require a lot more testing :slightly_frowning_face:.

This commit shouldn't change behavior apart from the logging.


Conan-Kudo:
@lukash Can you please rebase this?

PR #166, opened 2019-09-05
Fix: Verification of checksum from file attr

jrohel:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1700341

File copy could result in change in file attributes where
null-terminators are stripped out. The new code does not relly on it.

### DIFF
--- a/librepo/checksum.c
+++ b/librepo/checksum.c
@@ -221,18 +221,20 @@ lr_checksum_fd_compare(LrChecksumType type,
         // Load cached checksum if enabled and used
         struct stat st;
         if (fstat(fd, &st) == 0) {
-            ssize_t attr_ret;
             _cleanup_free_ gchar *key = NULL;
             char buf[256];
 
             key = g_strdup_printf("user.Zif.MdChecksum[%llu]",
                                   (unsigned long long) st.st_mtime);
-            attr_ret = fgetxattr(fd, key, &buf, 256);
-            if (attr_ret != -1) {
+            ssize_t attr_size = fgetxattr(fd, key, &buf, sizeof(buf));
+            if (attr_size != -1) {
                 // Cached checksum found
                 g_debug("%s: Using checksum cached in xattr: [%s] %s",
                         __func__, key, buf);
-                *matches = strcmp(expected, buf) ? FALSE : TRUE;
+                size_t expected_len = strlen(expected);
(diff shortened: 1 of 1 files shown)


j-mracek:
:+1: 
LGTM

librepo/checksum.c
```diff
@@ -221,18 +221,18 @@ lr_checksum_fd_compare(LrChecksumType type,
...
-            attr_ret = fgetxattr(fd, key, &buf, 256);
-            if (attr_ret != -1) {
+            ssize_t attr_size = fgetxattr(fd, key, &buf, sizeof(buf));
+            if (attr_size != -1) {
                 // Cached checksum found
                 g_debug("%s: Using checksum cached in xattr: [%s] %s",
                         __func__, key, buf);
-                *matches = strcmp(expected, buf) ? FALSE : TRUE;
+                size_t expected_len = strlen(expected);
+                *matches = expected_len > attr_size || memcmp(expected, buf, expected_len) ? FALSE : TRUE;
```
m-blaha: But this would return true whenever attribute value read from the filesystem starts with expected value. Isn't that too soft?

j-mracek: According to discussion with jrohel it is evaluated like:
(expected_len > attr_size || memcmp(expected, buf, expected_len)) ? FALSE : TRUE;
if expected_len > attr_size or data differs it returns FALSE.

m-blaha: I meant this situation:
expected='AB'
buf='ABCD'
Wouldn't it return TRUE then?

j-mracek: Yes, but the checksum has always the same length, therefore there is no chance of that situation. The problem here is that fgetxattr could theoretically return checksum with or without null character on the end (null character is not kept after file copy) therefore comparison of size could fail for no good real reason.

jrohel: @m-blaha 
Actual state:
The "expected_len" is 64 (sha256 encoded in hexadecimal characters string). The "attr_size" is 64 or 65 (depends if null terminator is present in attr).
But I agree with you. It is soft for general usage. I will change the condition to be more strict.

jrohel:
@m-blaha The comparison is strict now.

m-blaha:
@rh-atomic-bot r+

PR #168, opened 2019-09-11
misc: librepo/*.{c,h}: typo fixes.

Ionic:
Should, at most, change log output and error messages, otherwise be
non-functional.

One typo ("connectime") in the cache file format is not easily fixable
since that would have meant upgrading the cache version, but doing that
just for fixing a typo sounds stupid. Hence, let's leave a note for
later processing and hope for the best/that someone will actually
remember or read the note and act accordingly...

### DIFF
--- a/librepo/fastestmirror.c
+++ b/librepo/fastestmirror.c
@@ -38,6 +38,8 @@
 
 #define CACHE_GROUP_METADATA    ":_librepo_:"   // Group with metadata
 #define CACHE_KEY_TS            "ts"            // Timestamp
+// FIXME: next time the cache version is updated, please fix this typo! It
+// should obviously have been "connecttime".
 #define CACHE_KEY_CONNECTTIME   "connectime"    // Time of response
 #define CACHE_KEY_VERSION       "version"       // Version of cache format
 
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -112,12 +112,12 @@ typedef struct {
         Mirror */
     int allowed_parallel_connections; /*!<
         Maximum number of allowed parallel connections to this mirror. -1 means no limit.
-        Dynamicaly adjusted (decreased) if no fatal (temporary) error will occur. */
+        Dynamically adjusted (decreased) if no fatal (temporary) error will occur. */
     int max_tried_parallel_connections; /*!<
         The maximum number of tried parallel connections to this mirror
         (including unsuccessful). */
     int running_transfers; /*!<
-        How many transfers from this mirror are currently in progres. */
+        How many transfers from this mirror are currently in progress. */
     int successful_transfers; /*!<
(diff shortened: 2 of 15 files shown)


librepo/fastestmirror.c
```diff
@@ -39,6 +39,11 @@
 #define CACHE_GROUP_METADATA    ":_librepo_:"   // Group with metadata
 #define CACHE_KEY_TS            "ts"            // Timestamp
 #define CACHE_KEY_CONNECTTIME   "connectime"    // Time of response
+                                                // FIXME: next time the version
+                                                //        is updated, please
+                                                //        fix this typo! It
+                                                //        should obviously have
+                                                //        been "connecttime".
```
lukash: I'm thinking we should probably add a short reference to where the `SOVERSION` is defined in `librepo/CMakeLists.txt`, because otherwise this could be easily missed.

Ionic:
> Thanks for this! Out of curiosity, did you run a spellchecker against the sources?

Yep. During me packaging dnf for Debian (again, actually just updating the old packages I made a while ago), the QA tool lintian that also has checks for common spelling mistakes (but is not a fully-fledged spell checker) issued a few informational messages regarding this so I decided to fix them and looked around for some sort of source code spell checking tool. I ran across [scspell](https://github.com/myint/scspell) which caught the other typos, too.

I have to admit that it's generating a lot of noise, though, and has to be trained first by marking *a lot* of stuff as correctly written (like common abbreviations in code, function name parts or the like). That is, during the first run, it'll mark pretty much everything as misspelled.

librepo/fastestmirror.c
Ionic: I've been thinking about a way to make this more prominent, but I'm afraid that the cache structure version isn't actually coupled to the SOVERSION. The former has never been bumped up either so far.

I guess incrementing it with the next SOVERSION bump won't hurt, though, so why not.

lukash: Ah, sorry, I wasn't paying enough attention, I thought it's on the binary interface, but it is not, it is stored in the cache file. Unrelated to `SOVERSION`...

So yeah... looking at the code it probably wouldn't even break anything if we just changed it, as the cache would  simply be ignored (and I assume rewritten with the new key). But it's probably not enough reason to do it to fix the typo. (@j-mracek, @jrohel thoughts?)
If we won't be fixing it, could I ask to put the FIXME comment on the line above the constant and keep it a single line to make it a bit more subtle? :slightly_smiling_face:

Thanks!

Ionic: Right. I mean, I don't care. Updating the cache version means literally changing one digit. I just don't think that it makes a lot of sense. Also, it's just a stupid typo no user is likely to see, so we don't have to over-engineer stuff here.

Updated the comment/moved it above.

PR #173, opened 2019-10-03
Added some code using CPack for creation of deb packages

KOLANICH:


Conan-Kudo:
Sorry, I'm not adding CPack stuff. If you want deb packages, I can look at providing support in the spec we use to build RPMs.

KOLANICH:
>If you want deb packages, I can look at providing support in the spec we use to build RPMs.

Would be fine.

PR #178, opened 2019-11-27
Create a directory for gpg sockets in /run/user/ (RhBug:1769831,1771012)

lukash:
The solution of sending the "KILLAGENT" message to gpgagent to make it
clean up its sockets in gpg home dir is causing a race condition with
the gpgme_release() function.

Instead of trying to make the agent clean up its sockets (which doesn't
seem to be reliably possible), take advantage of its feature to create
the sockets under '/run/user/$UID' if this directory is present. The
sockets shouldn't be causing any trouble in this directory.

The commit creates the '/run/user/$UID' directory if it's not present on
the system. The sockets are then created there.

https://bugzilla.redhat.com/show_bug.cgi?id=1769831
https://bugzilla.redhat.com/show_bug.cgi?id=1771012

Fixes https://github.com/rpm-software-management/microdnf/issues/50

### DIFF
--- a/librepo/gpg.c
+++ b/librepo/gpg.c
@@ -32,28 +32,33 @@
 #include "util.h"
 #include "gpg.h"
 
-static void
-kill_gpg_agent(gpgme_ctx_t context, const char *home_dir)
-{
-    gpgme_error_t gpgerr;
-
-    gpgerr = gpgme_set_protocol(context, GPGME_PROTOCOL_ASSUAN);
-    if (gpgerr != GPG_ERR_NO_ERROR) {
-        g_warning("%s: gpgme_set_protocol: %s", __func__, gpgme_strerror(gpgerr));
-        return;
+/*
+ * Creates the '/run/user/$UID' directory if it doesn't exist. If this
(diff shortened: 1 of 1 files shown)


librepo/gpg.c
```diff
@@ -32,28 +32,33 @@
...
+ *
+ * [1] https://bugzilla.redhat.com/show_bug.cgi?id=1650266
+ * [2] https://bugzilla.redhat.com/show_bug.cgi?id=1769831
+ * [3] https://github.com/rpm-software-management/microdnf/issues/50
+ */
+void ensure_socket_dir_exists() {
+    char dirname[32];
+    snprintf(dirname, sizeof(dirname), "/run/user/%u", getuid());
+    int res = mkdir(dirname, 0700);
+    if (res != 0) {
```
rolweber: What happens if the directory already exists? Should you check for `res == EEXISTS` here?
http://man7.org/linux/man-pages/man2/mkdir.2.html

lukash: Right, thanks for noticing! I've taken it out to test logging of the message and forgot to put it back...

PR #181, opened 2020-01-18
Cleanup error messages

jdieter:
Make sure error messages include zchunk error message, including when write
callbacks fail

RHBZ: 1790625

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -1221,7 +1221,7 @@ find_local_zck_chunks(LrTarget *target, GError **err)
             }
 
             if(!zck_copy_chunks(zck_src, zck)) {
-                g_warning("Error copying chunks from %s to %s", cf, uf);
+                g_warning("Error copying chunks from %s to %s: %s", cf, uf, zck_get_error(zck));
                 zck_free(&zck_src);
                 close(chk_fd);
                 continue;
@@ -1334,7 +1334,7 @@ check_zck(LrTarget *target, GError **err)
         if(!cks_good) { // Error while validating checksums
             g_set_error(err, LR_DOWNLOADER_ERROR, LRE_ZCK,
                         "%s: Error validating zchunk file: %s", __func__,
-                        target->target->path);
+                        zck_get_error(zck));
             return FALSE;
         }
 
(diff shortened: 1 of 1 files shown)


librepo/downloader.c
```diff
@@ -1334,7 +1334,7 @@ check_zck(LrTarget *target, GError **err)
         if(!cks_good) { // Error while validating checksums
             g_set_error(err, LR_DOWNLOADER_ERROR, LRE_ZCK,
                         "%s: Error validating zchunk file: %s", __func__,
-                        target->target->path);
+                        zck_get_error(zck));
```
jdieter: We already log the filename further up the stack, so the actual error message is more important here.

librepo/downloader.c
```diff
@@ -1356,7 +1356,7 @@ check_zck(LrTarget *target, GError **err)
...
```
jdieter: We already log the filename further up the stack, so the actual error message is more important here.

librepo/downloader.c
```diff
@@ -1836,6 +1836,25 @@ check_finished_transfer_status(CURLMsg *msg,
...
+                g_info("Serious zchunk error: %s",
+                   zck_get_error(zck));
+                *serious_error = TRUE;
+            } else { // zck_is_error(zck) == 2
+                // Fatal zchunk error
+                g_info("Fatal zchunk error: %s",
+                   zck_get_error(zck));
+                *fatal_error = TRUE;
+            }
+            return TRUE;
```
jdieter: The above change will work with any version of zchunk, but zchunk versions < 1.1.5 will give somewhat cryptic error messages.

jdieter:
Tested by creating a painstakingly sized disk image designed to run out of disk space, mounting it to /var/cache/dnf, populating it with outdated metadata, running dnf update, and checking the error messages.

librepo/downloader.c
jdieter: FWIW, I completely agree that any error messages should include the file name.  In my original version of this code, the filename was actually listed twice.  `zck_get_error()` doesn't contain the file name (since the `zck` struct doesn't store that information), but whichever of these errors gets triggered will be appended to the error message in line 1514 which also includes the filename.

Sample message:
```Error: Failed to download metadata for repo 'updates': Yum repo downloading error: Downloading error(s): repodata/8b1f8fa78ab3f87e20e0e3a9a473fc23c303278947be95065e4845da17106dd6-filelists.xml.zck - Download failed: Zchunk error: Error writing data: No space left on device```

lukash: Oh, cool, sorry I've missed that.

jdieter: No problem at all.  It's nice to know you're looking closely enough at the PR to ask these questions.

PR #190, opened 2020-04-23
Fix memory leak in lr_check_repomd_xml_asc_availability()

m-blaha:
If signature downloading failed, `path` variable was not freed and file
descriptor was not closed.

### DIFF
--- a/librepo/metadata_downloader.c
+++ b/librepo/metadata_downloader.c
@@ -326,7 +326,7 @@ process_repomd_xml(GSList *targets,
         int fd_value = *((int *) fd->data);
 
         if (!target->handle || fd_value == -1) {
-            continue;
+            goto fail;
         }
 
         handle = target->handle;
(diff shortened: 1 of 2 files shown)


librepo/yum.c
```diff
@@ -476,6 +476,8 @@ lr_check_repomd_xml_asc_availability(LrHandle *handle,
             g_clear_error(&tmp_err);
             unlink(signature);
             lr_free(signature);
+            lr_free(path);
+            close(fd);
```
lukash, line 480: This doesn't seem right. Neither `fd` nor `path` is allocated/opened in this function, `fd` is not even used in this function besides closing it. This goes against common conventions of responsibility for deallocating. Whatever is passing these values to these functions might still be using them after this is called.

Having a look from where `lr_check_repomd_xml_asc_availability()` is called, one place is here: https://github.com/rpm-software-management/librepo/blob/e422278e4f0c505334aa109b1e1eb4ab7398ac3b/librepo/metadata_downloader.c#L341

There it's passing a `GSList` items as both `path` and `fd`. Unless I'm missing something, you are leaving a dangling pointer by freeing the list item this way. Having a quick look at the code it looks like a huge mess. The lists seem to be never freed, so it leaks as well. I think it needs to be fixed too...

For the other call to `lr_check_repomd_xml_asc_availability()` at: https://github.com/rpm-software-management/librepo/blob/e422278e4f0c505334aa109b1e1eb4ab7398ac3b/librepo/yum.c#L1312

Both variables should be freed (closed for `fd`) inside the fail-check if.

m-blaha: You are totally right. I tried to fix the leak with a minimal change of code and logic. I was also surprised by freeing allocated variable in other function... So I'll try to come up with a better fix.

librepo/metadata_downloader.c
```diff
@@ -421,6 +421,10 @@ lr_download_metadata(GSList *targets,
     }
 
     process_repomd_xml(targets, fd_list, paths);
+
+    g_slist_free(fd_list);
+    g_slist_free(paths);
```
lukash, line 426: I'm afraid it isn't that simple :confused: I think it should certainly go into a separate commit and...

There's `g_slist_free()` and `g_slist_free_full()`. The former function frees only the list itself and not its contents. But the contents of the `paths` list certainly aren't freed in every case inside `process_repomd_xml()`. The descriptiors in `fd_list` also aren't always closed. I haven't really carefully considered all the possible cases, it obviously depends on the path the code takes inside the `for` loop. Maybe the errors don't normally happen, but still. To me it seems the first three `continue`s leak both fds and paths.

This code structure is really unfortunate... My idea of fixing this is to put a `cleanup:` label to the end of the `for` loop in `process_repomd_xml()`, which would contain closing the fd and freeing the path. And then instead of the `continue`s `goto cleanup`. Or something along those lines...

m-blaha: The structure indeed is unfortunate but I'm not gonna fix this (at least in this PR). I tried to deal with memory / fd leaks in a better way, so please re-review it.

lukash:
Looking good now, thanks.

PR #197, opened 2020-08-19
Check if record->location_href contains .. before creating a path

loreson:
= changelog =
msg:           sanitize paths in prepare_repo_download_std_target and prepare_repo_download_zck_target
type:        security
resolves:      https://bugzilla.redhat.com/show_bug.cgi?id=1866498

### DIFF
--- a/librepo/yum.c
+++ b/librepo/yum.c
@@ -669,6 +669,17 @@ prepare_repo_download_std_target(LrHandle *handle,
                                  GSList **targets,
                                  GError **err)
 {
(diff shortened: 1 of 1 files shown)


kontura:
Thank you for the fix. 
Now this is kind of awkward because we already had a solution https://github.com/rpm-software-management/librepo/pull/198 but I was too slow in creating the PR, my bad.
I would be inclined to pick our version since it seems to me more robust and a bit more succinct, what do you think? :slightly_smiling_face:

loreson:
yes, your pr seems to be clearly superior,

PR #199, opened 2020-09-18
librepo: Fix the key string parsing in url_substitution

jaankit:
- Break the loop  if last char of key string is '_'
- Use Case:
  Lets say: releasever = 1.0
            basearch = x86_64
  URL String: $releasever_$basearch
  - Without fix:
        - parsed as $releasever_x86_64
  - With fix:
        - parsed as 1.0_x86_64

= changelog =
 msg: Fix the key string parsing in url_substitution
 type: bugfix
 resolves: string parsing issue in url_substitution

Signed-off-by: Ankit Jain <ankitja@vmware.com>

### DIFF
--- a/librepo/url_substitution.c
+++ b/librepo/url_substitution.c
@@ -124,7 +124,7 @@ lr_url_substitute(const char *url, LrUrlVars *list)
                 bracket = FALSE;
             }
             const char *varname = cur;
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@jaankit Thanks for the fix!

PR #203, opened 2020-10-12
cmake: mark libcheck as required for tests

lucab:
This adds an explicit `REQUIRED` annotation for libcheck when configuring
tests, aborting configuration if it cannot be found.


lukash:
Thanks!

lucab:
@lukash thank you for your prompt feedback! I'm still seeing some weird linkage issues (possibly due to a static `libcheck`), I may further follow up later on if once I figure out what's going on.

PR #204, opened 2020-10-13
cmake: allow building without Python bindings

lebauce:


lukash:
Looking good, thanks.

PR #205, opened 2020-10-15
cmake/tests: add libcheck linkage dependencies

lucab:
This tweaks tests linkage in order to bring in transitive dependencies
for `libcheck`. It fixes linking issues when building tests on systems
with a static `libcheck` library.


lucab:
Followup to https://github.com/rpm-software-management/librepo/pull/203. I've checked that this builds fine in (at least) the following buildroots:
 * deb, os-release -> "Debian GNU/Linux bullseye/sid"
 * rpm, os-release -> "Fedora 33 (Container Image Prerelease)"

lucab:
For context, the build failure is pretty much the same as shown in https://github.com/libcheck/check/issues/60 and the cmake fix in here is also modeled after those additional linking flags.

lukash:
To me it seems the transitive dependencies shouldn't be part of our CMakeLists.txt. `libcheck` uses pkg-config, see the output of `pkg-config --libs check` and `pkg-config --libs --static check`. The `PKG_CHECK_MODULES` macro should work for this.

lucab:
@lukash ah, you are absolutely right, I did not realize the .pc file was already covering all cases. I've amended the PR to use that, how does it looks now?

CMakeLists.txt
```diff
@@ -88,7 +88,7 @@ ENDIF (ENABLE_PYTHON)
 ADD_SUBDIRECTORY (librepo)
 
 IF (ENABLE_TESTS)
-  FIND_LIBRARY(CHECK_LIBRARY NAMES check REQUIRED)
+  PKG_CHECK_MODULES(LIBCHECK REQUIRED IMPORTED_TARGET check)
```
lukash, line 91: Is it necessary to use the `IMPORTED_TARGET` feature? I see it was added in CMake 3.6, the file claims minimum required version 2.8.5 and we should stick with that, maintaining compatibility with RHEL 7.

lucab: That's what makes the `PkgConfig::LIBCHECK` below working. I'll try to switch to some manual glueing to get rid of that.

tests/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ SET (librepotest_SRCS
 ADD_EXECUTABLE(test_main ${librepotest_SRCS})
 TARGET_LINK_LIBRARIES(test_main
     librepo
-    ${CHECK_LIBRARY}
+    ${LIBCHECK_LDFLAGS}
```
lukash, line 27: Ok, just one minor thing, we use the `_LIBRARIES` variables in other places (i.e. `LIBCHECK_LIBRARIES`), `TARGET_LINK_LIBRARIES` supports both formats. Can you change this for consistency sake? I've tested it, works on my end. Thanks!

lucab: I was still cross-checking this across distros, but `_LIBRARIES` does not seem to work at least on Debian:
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/10/../../../x86_64-linux-gnu/libcheck_pic.a(check_pack.o): undefined reference to symbol '__pthread_unregister_cancel@@GLIBC_2.3.3'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

Even if the .pc file there looks fine:
```
Requires: libsubunit
Libs: -L${libdir} -lcheck_pic  -pthread  -lrt -lm 
Libs.private:
Cflags: -I${includedir} -pthread
```

lucab: I guess the `-pthread`  is the one to blame,as `${LIBCHECK_LIBRARIES}` plus `${LIBCHECK_LDFLAGS_OTHER}` seem to work on both Centos7 and Debian buildroots. I've amended it that way, but let me know if you prefer some other arrangement.

lukash: Yeah, I was just composing a reply along those lines, though I can't reproduce your issue on Fedora with dymanic linking. I wonder what's the correct usage of these variables, but it seems the `_LDFLAGS` variable should be used here, as it contains all the required flags in one. Sorry about the hassle, could you put it back? :slightly_smiling_face: Thank you.

lucab: No worries at all, you were actually too quick and caught me while still in the middle of testing my changes :smile:. Amended.

lukash:
Thanks @lucab

PR #206, opened 2020-10-26
Various fixes for mirrors without ranges support and zchunk

kontura:
@jdieter could you please take a look at these changes if they make sense to you?

Fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1886706
Also possibly related to:
https://bugzilla.redhat.com/show_bug.cgi?id=1890434

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -1306,6 +1306,8 @@ check_zck(LrTarget *target, GError **err)
     if(target->mirror->max_ranges == 0 || target->mirror->mirror->protocol != LR_PROTOCOL_HTTP) {
         target->zck_state = LR_ZCK_DL_BODY;
         target->target->expectedsize = target->target->origsize;
+        free(target->target->range);
+        target->target->range = NULL;
         return TRUE;
     }
 
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -2293,11 +2293,7 @@ check_transfer_statuses(LrDownload *dd, GError **err)
                     if(zck_failed_chunks(zck) == 0 && zck_missing_chunks(zck) == 0)
                         target->zck_state = LR_ZCK_DL_FINISHED;
                 } else {
-                    if(target->range_fail) {
-                        target->range_fail = FALSE;
-                    } else {
-                        target->zck_state = LR_ZCK_DL_FINISHED;
-                    }
+                    target->zck_state = LR_ZCK_DL_FINISHED;
                 }
             }
             if(target->zck_state == LR_ZCK_DL_FINISHED) {
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -746,6 +746,10 @@ select_suitable_mirror(LrDownload *dd,
                 continue;
             }
 
+            if (mirrors_iterated == 0 && c_mirror->mirror->protocol == LR_PROTOCOL_FTP && target->target->is_zchunk) {
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


librepo/downloader.c
```diff
@@ -489,7 +489,7 @@ lr_headercb(void *ptr, size_t size, size_t nmemb, void *userdata)
     }
 
     #ifdef WITH_ZCHUNK
-    if(lrtarget->target->is_zchunk && lrtarget->mirror->max_ranges > 0)
+    if(lrtarget->target->is_zchunk && lrtarget->mirror->max_ranges > 0 && lrtarget->mirror->mirror->protocol != LR_PROTOCOL_FTP)
```
m-blaha, line 492: I didn't investigate it thoroughly so I might be wrong. Zchunk is usable only for protocols supporting ranges - HTTP(S) at the moment. Shouldn't be this test positive - i.e. `if (... && does_protocol_support_ranges(lrtarget->mirror->mirror->protocol))` (or even `&& protocol == LR_PROTOCOL_HTTP`). I'm basically thinking about other protocols librepo can handle (like rsync) which are not fixed by this patch.

jdieter: Zchunk does only work with http and https.  This would probably be better with just `&& protocol == LR_PROTOCOL_HTTP` (assuming that also covers https).

librepo/downloader.c
```diff
@@ -1303,6 +1303,9 @@ check_zck(LrTarget *target, GError **err)
     assert(!err || *err == NULL);
     assert(target && target->f && target->target);
 
+    /* Reset range fail flag */
+    target->range_fail = FALSE;
+
```
jdieter, line 1312: I'd like to take a closer look at the logic for this commit, since it's not clear to me what this is changing.

jdieter: Ok, I've looked more closely at this and I think it makes sense.  I've also rebuilt `librepo` with these commits added and can verify that it doesn't seem to break anything (I tested against an FTP mirror and a mirror without range support)

I am a bit concerned that I wasn't able to reproduce the bug https://bugzilla.redhat.com/show_bug.cgi?id=1886706 on my F33 system, either with stock librepo-1.12.1 or with this PR applied, so I'm not sure whether this PR actually fixes the problem.

librepo/downloader.c
kontura: Ok I have switched all the conditions to not be FTP specific but rather HTTP specific. (LR_PROTOCOL_HTTP also covers https)

librepo/downloader.c
kontura: Thanks for looking into this. I have actually changed it a bit more and added a new commit. I have tried to explain it in the commit messages but since its quite complex I am not sure they are readable.

That is weird I was able to reproduce it quite easily. Just having a zchunk repo served over FTP was enough. I have created tests for out ci-dnf-stack: https://github.com/rpm-software-management/ci-dnf-stack/pull/903 which also reproduce it (specifically the scenario with the `@bz1886706` tag).

lukash: @jdieter would you please review the latest changes? They are quite involved indeed...

jdieter (approved):
This all looks good to me.  The updated commit messages make the purpose of the changes (especially in the latter part of check_transfer_statuses) more clear.  I have also built the code and it seems to be working correctly on my 3G connection.

Conan-Kudo (approved):

lukash:
Thanks @jdieter (and @Conan-Kudo :slightly_smiling_face:)!

PR #208, opened 2020-11-06
fastestmirror: Fix memory leak

tinez:
I ran clang's 'scan-build' on the code base and it found a potential memory leak.

If any of the two 'curl_easy_setopt' calls in the loop fails, we'll leak the recently created 'mirror' object. This object is not really needed until the moment of appending it to 'list'. This patch delays the creation of the 'mirror' object, so there won't be a code path that leaks it.


m-blaha (approved):
Thanks!

PR #211, opened 2020-11-14
Sync data before writing checksum xattr

malmond77:
Writes to extended attributes are considered metadata, so can be commited to storage before data is fully synced. The upshot of this is that the checksum is cached but the file could be truncated. We attempt to sync data first to mitigate this problem.

This has been manifesting errors on real hosts with `package {p} does not verify: Payload SHA256 digest: BAD (Expected {a} != {b})` because `%_pkgverify_level` = `digest`. On inspection the files are nearly full size, the xattr is present as if the full file were there. The mtime in the xattr matches the file.

### DIFF
--- a/librepo/checksum.c
+++ b/librepo/checksum.c
@@ -250,6 +250,12 @@ lr_checksum_fd_compare(LrChecksumType type,
 
     *matches = (strcmp(expected, checksum)) ? FALSE : TRUE;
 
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

Conan-Kudo:
Merged, thanks for the patch!

PR #217, opened 2020-12-10
Fix: lr_perform() - Avoid 100% CPU usage

jrohel:
`curl_multi_wait()` behavior is strange.

libcurl documentation:
"If no extra file descriptors are provided and libcurl has no file descriptor to offer to wait for, this function will return immediately."

What means "libcurl has no file descriptor to offer to wait for"?
Does this mean that libcurl has no file descriptors? No.
It can have file descriptors. However, it can ignore them when the bandwidth limit is activated. If libcurl considers the bandwidth to be exceeded, `curl_multi_wait()` is not interested in these file descriptors (for the moment), so there are no file descriptors to wait for and `curl_multi_wait()` will return immediately.
In this case, there was a busy-loop in librepo and the processor was used to 100%.

Libcurl solved the problem by introducing new function `curl_multi_poll()` in version 7.66.0 . Because librepo must to work with the older libcurl, this fix implements an alternative solution.

https://bugzilla.redhat.com/show_bug.cgi?id=1878072

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -25,6 +25,7 @@
 #include <glib.h>
 #include <assert.h>
 #include <stdio.h>
+#include <time.h>
 #include <unistd.h>
 #include <string.h>
 #include <errno.h>
@@ -51,7 +52,6 @@
 #include "url_substitution.h"
 #include "yum_internal.h"
 
-#include <unistd.h>
 
 volatile sig_atomic_t lr_interrupt = 0;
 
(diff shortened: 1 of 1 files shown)


jrohel:
I sent new commit. It uses `nanosleep ()` instead of the deprecated `usleep ()`.

kontura:
@rh-atomic-bot r+

kontura:
oh there is no bot here, I will run the tests locally.

kontura:
Tests pass locally and LGTM.

PR #222, opened 2020-12-27
Add support for rpm2extents transcoder

malmond77:
Two related parts:

1. If `LIBREPO_TRANSCODE_RPMS` environment is set to a program (with parameters) then downloads are piped through it.
2. Transcoded RPMS by definition will not have the same bits on disk as downloaded. This is inherent. The transcoder is tasked with measuring the bits that enter stdin and storing a copy of the digest(s) seen in the footer. `librepo` can then use these stored digests instead if the environment variable is set.

This is part of changes described in https://fedoraproject.org/wiki/Changes/RPMCoW


librepo/checksum.c
```diff
@@ -101,6 +104,114 @@ lr_checksum_type_to_str(LrChecksumType type)
...
+        if (algo == NULL) {
+            g_set_error(err, LR_CHECKSUM_ERROR, LRE_TRANSCODE,
+                        "Cannot allocate %lu for algo", len + 1);
+            return NULL;
+
+        }
+        if (read(fd, algo, len) != len) {
+            g_set_error(err, LR_CHECKSUM_ERROR, LRE_TRANSCODE,
+                        "Cannot read algo");
+            free(algo);
```
jrohel, line 178: Resources should be allocated and deallocated by the same library.
When allocated with `calloc`, deallocation is done with `free`.
When allocated with `lr_malloc0`, deallocation must be done with `lr_free`!
The librepo API provides a deallocation function that should be used. Yes, it also works with `free`, because `lr_malloc` uses `calloc` internally. But it is wrong to rely on internal implementation, which can change.

It is repeated in several places.

librepo/checksum.c
```diff
@@ -101,6 +104,114 @@ lr_checksum_type_to_str(LrChecksumType type)
...
+        len = sizeof(algo_len_digest_len);
+        if (read(fd, &algo_len_digest_len, len) != len) {
+            g_set_error(err, LR_CHECKSUM_ERROR, LRE_TRANSCODE,
+                        "Cannot read algo_len_digest_len");
+            return NULL;
+        }
+
+        len = algo_len_digest_len.algo_len;
+        algo = lr_malloc0(len + 1);
+        if (algo == NULL) {
```
jrohel, line 169: The `lr_malloc0` uses `calloc` internally. That may change. Don't rely on it.

In addition, `lr_malloc` tests whether the memory was allocated. If memory is not allocated, `lr_out_of_memory` is called, which logs the problem and calls `abort`.

In other words, `lr_malloc` does not return `NULL`. If you need a function that returns `NULL` if the allocation fails, use `calloc`.

librepo/downloader.c
```diff
@@ -1443,6 +1448,132 @@ open_target_file(LrTarget *target, GError **err)
...
+                    "transcode env empty");
+        return;
+    }
+    if (pipe(transcoder_stdin) != 0) {
+        g_set_error(err, LR_DOWNLOADER_ERROR, LRE_TRANSCODE,
+                    "input pipe creation failed: %s",
+                    g_strerror(errno));
+        return;
+    }
+    if (fcntl(transcoder_stdin[1], F_SETFD, FD_CLOEXEC) != 0) {
```
jrohel, line 1485: I'm not sure if it makes sense to set the `FD_CLOEXEC` flag to `transcoder_stdin [1]` when the child process (line 1507) calls `close (transcoder_stdin [1]);`

malmond77: This warrants a comment: downloads can happen in parallel, and more than one transcoder can be run. I found that if I didn't close this pipe on exec, then the second transcoder ends up inheriting the write end. This means that the first trancoder doesn't get EOF as there's still a writer, and the thing deadlocks.

librepo/checksum.c
malmond77: Thanks, I will fix this, and the other places.

malmond77:
The second commit fixes an issue that only really surfaced when using public mirrors. In my previous testing (on CentOS) I didn't hit it because our internal mirrors were generally more consistent.

I'm aiming to get this code into Fedora 34 shortly, either as part of an updated tagged release (1.12.2?) or if not, just as a patch we carry in the rpm src. As part of the Fedora 34 change: this code path is intended to be optional. As far as I can tell, if the environment variable isn't set, then nothing should be different.

Conan-Kudo:
@malmond77 This looks good to me, but I am not comfortable merging this until the RPM PR is merged: https://github.com/rpm-software-management/rpm/pull/1470

DemiMarie:
@Conan-Kudo This is not a change I am comfortable with from a security perspective, as per past discussions.  The verification needs to happen before transcoding, which means either buffering the entire package on disk or changing the metadata format.

DemiMarie:
To elaborate: the idea of rpm2extents is fine, but it needs to be a separate entry in the metadata XML, with its own digest.

PR #228, opened 2021-02-03
Fix lr_yum_download_url in case lr_handle is NULL

m-blaha:
This simple python reproducer made librepo dump core:

```
>>> import os, librepo, tempfile
>>> t = tempfile.TemporaryFile()
>>> librepo.download_url('file:///etc/passwd', t.fileno())
Segmentation fault (core dumped)
```

### DIFF
--- a/librepo/yum.c
+++ b/librepo/yum.c
@@ -553,21 +553,23 @@ lr_yum_download_url(LrHandle *lr_handle, const char *url, int fd,
     gboolean ret;
     LrDownloadTarget *target;
     GError *tmp_err = NULL;
+    CbData *cbdata = NULL;
 
     assert(url);
     assert(!err || *err == NULL);
 
-    CbData *cbdata = cbdata_new(lr_handle->user_data,
-                                NULL,
-                                lr_handle->user_cb,
-                                lr_handle->hmfcb,
-                                url);
+    if (lr_handle != NULL)
+        cbdata = cbdata_new(lr_handle->user_data,
+                            NULL,
+                            lr_handle->user_cb,
+                            lr_handle->hmfcb,
+                            url);
 
     // Prepare target
     target = lr_downloadtarget_new(lr_handle,
                                    url, NULL, fd, NULL,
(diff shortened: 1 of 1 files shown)


librepo/yum.c
```diff
@@ -553,29 +553,32 @@ lr_yum_download_url(LrHandle *lr_handle, const char *url, int fd,
...
 
     assert(url);
     assert(!err || *err == NULL);
 
-    CbData *cbdata = cbdata_new(lr_handle->user_data,
-                                NULL,
-                                lr_handle->user_cb,
-                                lr_handle->hmfcb,
-                                url);
+    if (lr_handle)
```
lukash, line 561: So `lr_handle` is allowed to be NULL? Could you add a brief description of the case to the commit message? I could look it up myself, but you've likely already done the work and it's useful to have in the commit message for future reference.

m-blaha: Yes, the lr_handle is allowed to be NULL - according to documentation in python bindings that call this method - https://github.com/rpm-software-management/librepo/blob/a244d78bab11510408fb9413f1c78c2f56c69fa5/librepo/python/__init__.py#L1698 
I'll add the info to the commit message. I did not add curly brackets on purpose - to stay consistent with the coding style of the rest of librepo code. But I have no problem adding them.

lukash: > I did not add curly brackets on purpose - to stay consistent with the coding style of the rest of librepo code. But I have no problem adding them.

For the most part I don't like this argument. No big bang is ever going to happen wrt. changing these things (perhaps until the code is ultimately replaced as a whole). Hence gradual change seems the only way to actually apply the new style. It's case by case, seems fine here...

But leave it now, I'll test and merge :slightly_smiling_face: 

bors try

lukash:
bors retry

Conan-Kudo:
@m-blaha Can you rebase this and see if the GitHub Actions CI will be able to run through this?

m-blaha:
Oh, sure. I completely forgot this PR.

PR #232, opened 2021-03-01
Fix issue with re-downloading rpm package using reposync

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1931904

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/972

### DIFF
--- a/librepo/checksum.h
+++ b/librepo/checksum.h
@@ -107,6 +107,12 @@ lr_checksum_fd_compare(LrChecksumType type,
                        gchar **calculated,
                        GError **err);
 
+/** Clear all checksums cached in extended file attributes
+ * @param fd            File descriptor
+ */
+void
+lr_checksum_clear_cache(int fd);
+
 /** @} */
 
 G_END_DECLS
--- a/librepo/checksum.c
+++ b/librepo/checksum.c
@@ -269,3 +269,34 @@ lr_checksum_fd_compare(LrChecksumType type,
 
     return TRUE;
 }
+
+
+void
+lr_checksum_clear_cache(int fd)
+{
+    char *xattrs = NULL;
+    ssize_t xattrs_len;
+    ssize_t bytes_read;
+    const char *attr;
+    const char *prefix = "user.Zif.MdChecksum[";
+    ssize_t prefix_len = strlen(prefix);
+
+    xattrs_len = FLISTXATTR(fd, NULL, 0);
+    if (xattrs_len <= 0) {
+        return;
+    }
+    xattrs = lr_malloc(xattrs_len);
+    bytes_read = FLISTXATTR(fd, xattrs, xattrs_len);
+    if (bytes_read < 0) {
+        goto cleanup;
+    }
+    attr = xattrs;
+    while (attr < xattrs + xattrs_len) {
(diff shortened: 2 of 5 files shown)


librepo/checksum.h
```diff
@@ -107,6 +107,38 @@ lr_checksum_fd_compare(LrChecksumType type,
...
+ * @param calculated    If not NULL, the calculated checksum will be pointed
+ *                      here, the pointed string must be freed by caller.
+ * @param timestamp     Use this timestamp for checksum caching. If -1,
+ *                      timestamp is taken from file's mtime.
+ * @param err           GError **
+ * @return              returns TRUE if error is not set and FALSE if it is
+ */
+gboolean
+lr_checksum_fd_compare_mtime(LrChecksumType type,
+                       int fd,
```
kontura, line 127: I see you are still updating this so I didn't look into it too much but can you fix the alignment of arguments for `lr_checksum_fd_compare_mtime` here and also its definition has the same problem.

Additionally I think the `test_cached_checksum` unit test in test_checksum.c needs an update.

m-blaha: I fixed the indent issue and will look to the unit test stuff.

m-blaha: It looks like the `test_cached_checksum` was never working well - the test file is created in `/tmp/` which is usually a tmpfs filesystem. Even though tmpfs supports extended attributes, usage of the `user` xattrs (on which librepo's checksum caching relies) is not permitted here. (see man tmpfs)

kontura: I see, that would explain why the test was passing on my host machine (it was skipping the check because it was on tmpfs) but it was failing in podman.

m-blaha: I fixed the `test_cached_checksum` test and added new  tests for `lr_checksum_fd_compare_mtime` and `lr_checksum_clear_cache`.
I'm not sure if i should try to create temp files out of the /tmp/ directory - even then you can never be sure that the filesystem you are using supports them.

m-blaha:
@kontura sorry, the patch was originally over-complicated. There is no need to create new `lr_checksum_fd_compare_mtime` API so I've removed it.

kontura:
bors try

librepo/checksum.h
kontura: I agree, the current approach in the unittests is sufficient.

What about the CI tests? Will we have the same problem there? I tried searching for what is the status of xattrs in containers but didn't manage to find anything definitive. It does work in my podman runs.

kontura: Or it actually shouldn't matter from reposync's point of view, it will just compute the checksum again, but the behavior would be the same?

m-blaha: That's a good point actually. When the CI test runs on the tmpfs, the extended attributes are not used, the checksums are re-calculated and the re-downloads will not occur - the test will pass regardless the patch is present.
Although I do not have the CI test for the bug ready yet, from what I've seen the extended attributes are working in our test containers so I'll rely on that and prepare the test. I will add a step there checking that the extended attribute on the downloaded file is actually present - just to make it fail in case the xattrs are not supported.

kontura (approved):
LGTM, tests work.

Issue #233, opened 2021-03-11
lr_checksum_fd_compare() does not return cached calculated value


malmond77:
Context: I'm working on https://fedoraproject.org/wiki/Changes/RPMCoW

I've found that DNF still uses some old yum code to produce and validate checksums of downloaded rpms (https://github.com/rpm-software-management/dnf/blob/master/dnf/yum/misc.py#L153). In the normal case: librepo downloads and validates rpm files using librepo. In the less common path, the rpm is already there, and dnf uses the python code linked to verify the checksum.

This works today because the implementation is close enough to not matter.

Enter: #222 . This augments checksum handling with specific comprehension of "transcoded" rpms. This works in the normal path, but fails with existing packages in cache because the python code is used, and it doesn't know about the transcoded .recorded checksum.

Solving this is fairly involved. There's changes to librepo, libdnf, and dnf needed to fix this. As listed: librepo is first, and I'd like to fix a latent bug in the API, or get guidance on how to better do it.

My thought is to make dnf use `lr_checksum_fd_compare()` and `lr_checksum_fd_cmp()` instead of` yum.misc.checksum()`.  Why both? Because in one case, we need to validate checksums (https://github.com/rpm-software-management/dnf/blob/master/dnf/package.py#L333) and another we need to just calculate it (https://github.com/rpm-software-management/dnf/blob/master/dnf/package.py#L59)

The API for librepo doesn't explicitly support calculating checksums. The closest we get is is to call `lr_checksum_fd_compare()` and pass an empty string for `expected`, ignore `matches`, and pass a `char *` for `calculated`.

If you call `lr_checksum_fd_compare()` with `caching = TRUE`, and an xattr is present, then `matches` is set, but `calculated` is not set. This feels like a bug: if the caching mechanic is robust enough to use for the normal case, then it should be possible to set `*calculated` too. Using caching shouldn't come with this drawback.

For now, I'm going to call `lr_checksum_fd_compare()` in my `libdnf` SWIG wrapper with `caching = FALSE` to force a full checksum though librepo. I will reference this issue in my code, and one day, when the fix to this issue is available in downstream packages, I can enable caching for this less common path.

malmond77:
Fix is merged. I will update https://github.com/rpm-software-management/libdnf/pull/1164 to explicitly depend on v1.13.1 which has the fix so I can reliably pass TRUE for caching in `checksum_value()`. I will also bump versions and depend on them in dnf itself, completing the chain.

PR #234, opened 2021-03-11
Return "calculated" checksum if requested w/caching

malmond77:
If a file is downloaded via librepo (e.g. `dnf install --downloadonly`)
then a request to get the checksum via `lr_checksum_fd_compare()` will
not work. It'll only return whether the checksum is valid, and not the
actual checksum. This is the simple fix.

Addresses #233


librepo/checksum.c
```diff
@@ -248,6 +248,8 @@ lr_checksum_fd_compare(LrChecksumType type,
                     g_debug("%s: Using checksum cached in xattr: [%s] %s",
                             __func__, checksum_key, buf);
                     *matches = (strcmp(expected, buf) == 0);
+                    if (calculated)
+                      *calculated = g_strdup(checksum);
```
m-blaha, line 252: `checksum` is always NULL here. Did you mean to return the cached checksum value stored in `buf`?

malmond77: You're right: I published this without even building it (for shame). As penance and because it's a good idea, I've both fixed this silly mistake, and extended testing to formalize the use of `lr_checksum_fd_compare()` as a way to calculate or fetch checksums explicitly, which is the point of https://github.com/rpm-software-management/libdnf/pull/1164

tests/test_checksum.c
```diff
@@ -178,6 +178,103 @@ START_TEST(test_cached_checksum)
...
+    attr_ret = GETXATTR(filename, checksum_key, &buf, sizeof(buf)-1);
+
+    if (attr_ret == -1) {
+        // Error encountered
+        if (errno == ENOTSUP) {
+            // Extended attributes are not supported
+            goto exit_label;
+        }
+        // Any other errno means fail
+        fail_if(attr_ret == -1);
```
m-blaha, line 238: Testing of xattrs is tricky and two weaks ago I had tough times with it so I can show off a bit ;) . Long story short - the test will fail on this line in case the test file is created on regular filesystem - not tmpfs which /tmp/ usually is.

- xattrs are supported on tmpfs but user extended attributes are not permitted (see man tmpfs)
- so if you try `make test` on your machine the test will pass - because the test file is created inside the `/tmp/` dir which is probably a tmpf, you get ENOTSUP error on line 233 and the rest of the test is skipped.
- but if you create the test file on a regular filesystem (either by changing the location on the line 197 or by trying to build a rpm from this PR using mock - that is how I discovered the problem) you find out that the test doesn't work. Instead of `ENOTSUP` you start getting `ENODATA`.
- the reason is another bug in librepo - the computed checksum is written to the xattr only when expected and calculated checsum matched - see `librepo/checksum.c:274`. Once calculated the checksum should be written to xattr regardless it matches.

I didn't solve the issue with tests and tmpfs - it cannot be 100% solved but we probably  could minimize its impact by creating testing files somewhere inside the build directory - passing it to to `test_main` using current `run_tests.sh` the same way as the repos directory is passed? But this definitely might be done later in other PR.

m-blaha (changes requested):

tests/test_checksum.c
malmond77: This is a very good catch. Given the current semantics, I think it's better to omit the xattr checks (because we'll never have `expected == calculated`). I'm keen to get this minor change in, because it's objectively less wrong than before. I will update this test, and omit this incorrect code for now.

malmond77: See #235 for the caching bug. Note this isn't as simple as I originally thought.
#236 for the xattr issue 

If these are dupes of existing issues, please feel free to close them.

m-blaha:
Please, can you also bump the `Version:` field in librepo.spec to be in sync with VERSION.cmake? We are using this spec file to build our nightlies.
Other then this the PR is ready to be merged. Thanks for your work!

m-blaha:
Thanks! I'll wait for tests to finish and merge it.

m-blaha (approved):

PR #240, opened 2021-04-21
Truncate downloaded file more correctly

malmond77:
The previous code would only truncate the file, and turn off resume if the xattr was missing. This left a gap when resume was disabled, and the xattr was missing - the file wouldn't get truncated if present.

This behaviour is strictly needed for #222 to work correctly. That does not support resuming (sorry, one day it could), so it explicitly disables resuming.


malmond77:
I should have put this as a draft while I work on this, do not review yet :

PR #241, opened 2021-04-28
Replace python3-flask with http.server in python tests

pkratoch:


m-blaha:
LGTM and tests are working (actually they are even faster then before). Thanks!

PR #243, opened 2021-05-12
Reduce time to load metadata

papoteur-mga:
See https://github.com/rpm-software-management/libdnf/issues/1216
Replace `zck_validate_checksums()` with `zck_validate_data_checksum()` to avoid to calculate `SHA512` sums in excess of `SHA256`. 
An evaluation with callgrind says that `lr_yum_perform` takes only 22% of time instead of 40% when loading metadata.


kontura:
That makes sense, thank you for the patch and detailed investigation. :+1:

PR #244, opened 2021-05-12
Fix link to the online documentation in README.md

msehnout:
Github pages are available on the github.io domain instead of github.com. That's why the link didn't work. This patch fixes the link.


m-blaha:
Thanks!

PR #245, opened 2021-05-17
Recover from fsync fail on read-only filesystem (RhBug:1956361)

kontura:
When `fsync` fails due to the file not supporting synchronization just log
the problem instead of failing the whole dnf run. This happens for
example with filesystems mounted read-only in which case there is no
point to `fsync` anyway.

Currently we also ignore return values from `FSETXATTR` which also fails
on read-only filesystem (so no checksum cache is set). This is fine however
since the checksum is recomputed when needed, dnf is just a bit slower.

https://bugzilla.redhat.com/show_bug.cgi?id=1956361

@malmond77 can you take a look at this please?

### DIFF
--- a/librepo/checksum.c
+++ b/librepo/checksum.c
@@ -266,9 +266,13 @@ lr_checksum_fd_compare(LrChecksumType type,
     *matches = (strcmp(expected, checksum)) ? FALSE : TRUE;
 
     if (fsync(fd) != 0) {
-        g_set_error(err, LR_CHECKSUM_ERROR, LRE_FILE,
-                    "fsync failed: %s", strerror(errno));
-        return FALSE;
+        if (errno == EROFS || errno == EINVAL) {
+            g_debug("fsync failed: %s", strerror(errno));
+        } else {
+            g_set_error(err, LR_CHECKSUM_ERROR, LRE_FILE,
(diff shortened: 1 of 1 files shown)


remicollet:
I confirm this patch fix the issue 

Tested on Fedora 33 with librepo 1.14.0 + this patch and dnf 3.7.0 + mock 2.10.1

malmond77:
@kontura : this makes a lot of sense to me. Thanks for writing this!

m-blaha (approved):
LGTM, if you consider this to be ready (ready for review label is not applied), do not hesitate to merge it.

pghmcfc:
This bug is affecting me on Fedora 34 with dnf-4.7.0-1.fc34 and librepo-1.14.0-1.fc34; is there likely to be an update at some point to fix it?

kontura:
@pghmcfc there will definitely be an update to F34 but it might take a while.

kontura:
@pghmcfc there is https://bodhi.fedoraproject.org/updates/FEDORA-2021-3b74285c43 which contains new librepo with the fix.

pghmcfc:
> @pghmcfc there is https://bodhi.fedoraproject.org/updates/FEDORA-2021-3b74285c43 which contains new librepo with the fix.

Thanks, that works for me.

PR #246, opened 2021-05-25
Specify repomd.xml which is not redownloaded if not modified since its mtime

kontura:
This uses the `if-modified-since` header (via libcurl) so that we don't
have to redownload identical repomd.xml from servers that support this.

It can be used by specifying a path to existing repomd.xml in the librepo
handle. mtime of this repomd.xml is then used in the `if-modified-since`
header and if the server returns `304` (the file wasn't modified since)
we copy the current repomd.xml to the new target location which ensures
librepo and all the other tooling can continue as if it was downloaded.

= changelog =
msg: Use if-modified-since header to avoid redownload if path to exisiting repomd is specified
type: enhancement
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1855296


kontura:
For: https://github.com/rpm-software-management/libdnf/pull/1243

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1002

kontura:
We have decided not to continue with this, more info in the bz1855296.

PR #248, opened 2021-06-22
Covscan warnings and `fail_if` deprication (RhBug:1938782)

kontura:
Replacing all the `fail_if`'s makes this quite big and I would say hard to review since we are going from `fail_if` to `assert` so negation of the conditions is usually needed.

For: https://bugzilla.redhat.com/show_bug.cgi?id=1938782

### DIFF
--- a/tests/test_downloader.c
+++ b/tests/test_downloader.c
@@ -46,7 +46,7 @@ START_TEST(test_downloader_single_file)
     fail_if(handle == NULL);
 
     char *urls[] = {"http://www.google.com", NULL};
-    lr_handle_setopt(handle, NULL, LRO_URLS, urls);
+    fail_if(!lr_handle_setopt(handle, NULL, LRO_URLS, urls));
     lr_handle_prepare_internal_mirrorlist(handle, FALSE, &tmp_err);
     fail_if(tmp_err);
 
@@ -55,8 +55,7 @@ START_TEST(test_downloader_single_file)
 
     tmpfn1 = lr_pathconcat(test_globals.tmpdir, "single_file_XXXXXX", NULL);
 
-    mktemp(tmpfn1);
-    fd1 = open(tmpfn1, O_RDWR|O_CREAT|O_TRUNC, 0666);
+    fd1 = mkstemp(tmpfn1);
     lr_free(tmpfn1);
     fail_if(fd1 < 0);
 
@@ -86,6 +85,7 @@ START_TEST(test_downloader_single_file)
     }
 
     g_slist_free_full(list, (GDestroyNotify) lr_downloadtarget_free);
+    close(fd1);
 }
 END_TEST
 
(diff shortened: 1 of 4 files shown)


lukash:
Wow... very hard to review the last commit. I could do it locally, but what about removing the last commit from this PR and creating a special PR for it? That way if there are only the `fail_if` -> asserts changes and they all pass, we have reasonable certainty the change is correct (for binary checks we can be certain, for e.g. range checks not so much, but those aren't too many).

tests/test_downloader.c
```diff
@@ -55,7 +55,7 @@ START_TEST(test_downloader_single_file)
 
     tmpfn1 = lr_pathconcat(test_globals.tmpdir, "single_file_XXXXXX", NULL);
 
-    mktemp(tmpfn1);
+    fail_if(mkstemp(tmpfn1) == -1);
```
jrohel: * The original code uses `mktemp`. The `mktemp` function generates a unique temporary filename from template.
* The new code uses `mkstemp`. The `mkstemp()` function generates a unique temporary filename from template, **creates and opens the file, and returns an open file descriptor for the file** . The code compares the returned descriptor with -1 and discards it. **File descriptor leak**. On the next line, the code reopens an already open file - the `open` function. 

The problem occurs in many places.

kontura: I see, thank you.

I have changed it to use the file descriptor from the `mkstemp` call directly and added closing it at the end of tests.
(I have also tried using `_cleanup_fd_close_` which librepo defines but covscans were complaining about it.)

kontura:
> Wow... very hard to review the last commit. I could do it locally, but what about removing the last commit from this PR and creating a special PR for it? That way if there are only the `fail_if` -> asserts changes and they all pass, we have reasonable certainty the change is correct (for binary checks we can be certain, for e.g. range checks not so much, but those aren't too many).

ok, I have removed the last commit from here and will make a separate PR once this is merged (since I am adding few `fail_if`'s here).

tests/test_repoconf.c
```diff
@@ -519,8 +503,8 @@ START_TEST(test_write_repoconf)
     _cleanup_error_free_ GError *tmp_err = NULL;
 
     // Create a temporary file
-    rc = mkstemp(tmpfn);
-    fail_if(rc == -1);
+    fd = mkstemp(tmpfn);
+    fail_if(fd == -1);
```
lukash, line 507: I know it was like this already, but since `mkstemp` opens the file, maybe we should close it here too. Not that it matters much for the tests, but...

kontura: Here it is taken care of by `_cleanup_fd_close_` (at line 496), I am honestly not sure why the covscan is ok with it here but wasn't when I tried to use it in the other tests, maybe its because it isn't used here for anything.

lukash: Oh, I see. Looks good then.

PR #258, opened 2022-07-15
Fix alloc / free mismatches from covscan

lukash:


kontura:
This does resolve a bunch of mismatches but unless I am mistaken it seems that with these fixed the covscan is able to detect even more.
I did a run without this patch [here](https://cov01.lab.eng.brq.redhat.com/covscanhub/task/257303/) vs. run with the patch [here](https://cov01.lab.eng.brq.redhat.com/covscanhub/task/257304/). In the `san-results.html` we can see there is more new `Error: ALLOC_FREE_MISMATCH`es.

lukash:
I've fixed most of the additional covscan issues, barring a few "non-issues" which I believe shouldn't pop up in the real run.

kontura:
Looks good!
Thank you.

PR #261, opened 2022-08-16
Update libattr URL in README.md

markdirish:
Porting librepo for IBM i, ran into the libattr dependency. Current URL in the README timedout. Found the new URL for the project at: https://savannah.nongnu.org/projects/attr/

This is the URL used in both the [Fedora](https://src.fedoraproject.org/rpms/attr/blob/rawhide/f/attr.spec) and [Ubunutu](https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/libattr1-dev_2.4.48-5_amd64.deb.html) packages.


kontura:
Thanks!

PR #262, opened 2022-08-22
Detailed error message when using non-existing TMPDIR (RhBug:2019993)

jan-kolarik:
Show detailed info about missing temporary directory when `TMPDIR` directory was removed or the environmental variable was unset before its processing during `dnf` command.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2019993.


jrohel:
OK. Thanks for fix.

PR #263, opened 2022-08-27
metadata_downloader: Include unistd.h for lseek()

kraj:
This is found when compiling on musl systems

Fixes

metadata_downloader.c:331:9: error: call to undeclared function 'lseek'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        lseek(fd_value, SEEK_SET, 0);
        ^

Signed-off-by: Khem Raj <raj.khem@gmail.com>


kontura:
Thanks!

PR #264, opened 2022-08-31
Network wait API support

### LINKED ISSUE #259: Add API/support for waiting on network
Breaking this out from https://github.com/coreos/rpm-ostree/issues/3096

Basically I'm thinking we could add something like 

```
lr_handle_setopt(priv->repo_handle, error, LRO_NETWORK_WAIT, TRUE);
```

This would end up parsing the URL `baseurl` or `metalink` and using https://developer-old.gnome.org/gio/2.68/GNetworkMonitor.html to wait until the URL (host) was reachable.

A big question in my mind is - could we actually enable this by default?  It's tempting to try, but the "blast radius" seems large too.  So perhaps better to opt-in to start at least.

There's someone on the CoreOS team interested in working on this potentially; does this proposal make sense to you?

RishabhSaini:
Resolves #259 

handle.h: added option for LRO_NETWORK_WAIT and a helper function for handling netowork wait
handle.c: Added callback function required for GMainEventLoop and functionality for option
handle-py.c: Added the case for LRO_NETWORK_WAIT

Adds API support for waiting on network:
This waits for clients network interfaces till the specified time.
lr_handle_setopt(priv->repo_handle, error, LRO_NETWORK_WAIT, seconds)
This parses metalinkurl/baseurl and polls g_network_monitor_can_reach() in an event driven
manner with the help of GMainEventLoop.


librepo/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ TARGET_LINK_LIBRARIES(librepo
                         ${LIBCRYPTO_LIBRARIES}
                         ${GPGME_VANILLA_LIBRARIES}
                         ${GLIB2_LIBRARIES}
+                        ${GIO2_LIBRARIES}
```
m-blaha, line 55: `GIO2_LIBRARIES` variable is uninitialized, so the project is not linked with gio library. Thus the `/usr/bin/ld: ../librepo/librepo.so.0: undefined reference to g_network_monitor_get_default` linker error (see the copr build log in Check tab - https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/CI-librepo-pr264/build/4800193)

One solution - drop this line and change https://github.com/rpm-software-management/librepo/blob/66d4f81d6bc75215179ff2517be4338ee50cccb6/CMakeLists.txt#L31 line to
`PKG_CHECK_MODULES(GLIB2 glib-2.0>=2.28 gio-2.0 REQUIRED)`

librepo/handle.c
```diff
@@ -841,7 +867,57 @@ lr_handle_setopt(LrHandle *handle,
...
+    gchar *baseurl;
+    if(handle->metalinkurl)
+      baseurl = handle->metalinkurl;
+    else if(handle->mirrorlisturl)
+      baseurl = handle->mirrorlisturl;
+    else if(handle->urls)
+      baseurl = handle->urls[0];
+    assert(baseurl);
+
+    GUri *uri = g_uri_parse(baseurl, G_URI_FLAGS_NONE, NULL);
```
cgwalters, line 900: I think one thing we should to do here is call `g_uri_get_scheme()` and if it's `file` then we bypass all of this.

librepo/handle.c
```diff
@@ -841,7 +867,57 @@ lr_handle_setopt(LrHandle *handle,
...
+      return FALSE;
+    }
+    const gchar* host = g_uri_get_host(uri);
+    guint16 port = g_uri_get_port(uri);
+    GSocketConnectable *connectable =  g_network_address_new(host, port);
+    data_struct.connectable = connectable;
+    GMainLoop *loop;
+    loop = g_main_loop_new(NULL, FALSE);
+    data_struct.loop = loop;
+    g_timeout_add(200, timeout_callback, &data_struct);
```
cgwalters, line 911: Hmm.  I'd hoped there was a way with GNetworkMonitor to be event driven, but it seems like there's actually not.  That means that this use of a mainloop is not functionally different from the previous `while` loop you had.  (But, I guess it's slightly easier to read?)

That said, what we *can* do is monitor in an event-driven way the `network-available` boolean.  If that's `FALSE` then we can reasonably assume that no remote hosts are reachable.

But eh...it doesn't matter too much I guess.

librepo/handle.c
```diff
@@ -841,7 +867,57 @@ lr_handle_setopt(LrHandle *handle,
...
+    const gchar* host = g_uri_get_host(uri);
+    guint16 port = g_uri_get_port(uri);
+    GSocketConnectable *connectable =  g_network_address_new(host, port);
+    data_struct.connectable = connectable;
+    GMainLoop *loop;
+    loop = g_main_loop_new(NULL, FALSE);
+    data_struct.loop = loop;
+    g_timeout_add(200, timeout_callback, &data_struct); 
+    g_main_loop_run(loop);
+    g_main_loop_unref(loop);
```
cgwalters, line 913: We can use `g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, FALSE);` above, then drop this manual unref.

librepo/handle.c
```diff
@@ -812,6 +834,10 @@ lr_handle_setopt(LrHandle *handle,
         c_rc = curl_easy_setopt(c_h, CURLOPT_FILETIME, handle->preservetime);
         break;
 
+    case LRO_NETWORK_WAIT:
+        assert(lr_handle_network_wait(handle, err, va_arg(arg, long)));
```
cgwalters, line 838: Do we need this option at all anymore actually, if we're requiring the caller to invoke the API method directly?

librepo/handle.c
RishabhSaini: I am unable to comprehend why is the current implementation using GMainEventLoop similar to the "while" loop implementation. It actively waits on the g_network_monitor_can_reach till the network-timeout expires. Also to be able to monitor the network-available boolean in an event-driven manner, I would still need to call the g_network_monitor_can_reach and set boolean variable to true/false appropriately. So how do the two methods differ?

cgwalters: Right, but as I understand it `g_network_monitor_can_reach` will immediately return if the host is unreachable (e.g. if there's no default route).  Then 200 milliseconds later, we'll wake up and try again.

IOW, it's still polling - and that makes it similar to the `while` loop.

> Also to be able to monitor the network-available boolean in an event-driven manner, I would still need to call the g_network_monitor_can_reach and set boolean variable to true/false appropriately. 

Yes, but if I understand things correctly, if the machine is completely offline (e.g. cloud system that is still booting up), then the `network-available` boolean will be `FALSE` and we *know* that connections won't work.

But, very quickly after DHCP works and we get a route, then we'll get a signal from NetworkManager (or kernel netlink) and that will wake the process up.

Another way to look at this is: it's an optimization to avoid polling when there's no network.

(To be very clear: I *could be wrong* about some of the details here!)

RishabhSaini: Ok that makes sense. Now to implement:

> monitor in an event-driven way the network-available boolean

Option 1: I tried exploring creating custom events in the GMainEventLoop using `g_source_new` and using [GSourceFuncs](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/2012.5/glib/glib-The-Main-Event-Loop.html#GSourceFuncs) such as prepare, check, dispatch, finalize to implement the logic. An example of [Creating a new source (scroll down)](https://docs.huihoo.com/symbian/nokia-symbian3-developers-library-v0.8/GUID-7FD05006-09C1-4EF4-A2EB-AD98C2FA8866.html). These custom events are not time based and hence won't poll at a specific interval. The logic of callback function will almost remain the same, but the GSourceFuncs will be added.
I am not sure whether this is still method will still poll like a simple `while` loop implementation.

Option2: Is there any other method not involving GMainEventLoop at all?

cgwalters: Use `g_signal_connect("notify::network-available")` on the monitor instance, pass the main loop as user data.  When the `network-available` property becomes `TRUE`, you could exit the loop, and then add the timer to the loop and restart it.

librepo/handle.c
```diff
@@ -264,6 +266,33 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+   GNetworkMonitor *monitor;
+   GSocketConnectable *connectable; 
+   GCancellable *cancellable;
+};
+
+gboolean timeout_callback(gpointer data)
+{ 
+  struct callback_data *dt = (struct callback_data*)data;
+   if(!g_network_monitor_can_reach(dt->monitor, dt->connectable, dt->cancellable, NULL) &&
+      time(NULL) - dt->begin <= dt->input_seconds){
```
cgwalters, line 282: `time` operates in terms of wall clock time - this means it can change if e.g. the system clock is out of date and resynchronizes.

Further, when the system suspends (e.g. closing laptop lid) then the values can make *large* leaps.

We should use `g_get_monotonic_time()` here instead.  A bit more about [monotonic clocks](https://itnext.io/as-a-software-developer-why-should-you-care-about-the-monotonic-clock-7d9c8533595c).

librepo/handle.c
```diff
@@ -841,7 +870,56 @@ lr_handle_setopt(LrHandle *handle,
...
+    assert(!err || *err == NULL);
+
+    if (!handle) {
+        g_set_error(err, LR_HANDLE_ERROR, LRE_BADFUNCARG,
+                    "No handle specified");
+        return FALSE;
+    }
+
+    GNetworkMonitor *monitor = g_network_monitor_get_default();
+    GCancellable *cancellable = g_cancellable_new ();
```
cgwalters, line 886: What would be better here is to have this cancellable be a parameter to the API function.  As of right now, the cancellable here is doing nothing because nothing will ever cancel it.  If we make it part of the API, then e.g. the calling code can choose to cancel the wait.

librepo/handle.c
```diff
@@ -841,7 +870,56 @@ lr_handle_setopt(LrHandle *handle,
...
+      return TRUE; 
+    }
+    const gchar* host = g_uri_get_host(uri);
+    guint16 port = g_uri_get_port(uri);
+    GSocketConnectable *connectable =  g_network_address_new(host, port);
+    data_struct.connectable = connectable;
+    g_autoptr(GMainLoop) loop;
+    loop = g_main_loop_new(NULL, FALSE);
+    data_struct.loop = loop;
+    g_signal_connect(monitor, "notify::network-available", G_CALLBACK(on_network_available), &data_struct);
```
cgwalters, line 918: I think, what we need to do is query the property first here.  If the network *is* available at the moment then we should proceed directly to waiting on the URL.

librepo/handle.c
```diff
@@ -264,6 +266,33 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+  struct callback_data *dt = (struct callback_data*)data;
+   if(!g_network_monitor_can_reach(dt->monitor, dt->connectable, dt->cancellable, NULL) &&
+      time(NULL) - dt->begin <= dt->input_seconds){
+      return TRUE;
+   } 
+   g_main_loop_quit(dt->loop);
+   return FALSE;
+}
+
+void on_network_available(GObject *object, GParamSpec *pspec, gpointer data)
```
cgwalters, line 289: When we get notified about this, it may actually be because the property transitioned from `true` :arrow_right: `false`!  So...related to my comment below, I think to be a bit more precise here we should call `g_network_monitor_get_network_available ()` - and start/stop polling the URI based on that.

librepo/handle.c
```diff
@@ -264,6 +266,32 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+   GNetworkMonitor *monitor;
+   GSocketConnectable *connectable; 
+   GCancellable *cancellable;
+};
+
+gboolean timeout_callback(gpointer data)
+{ 
+  struct callback_data *dt = (struct callback_data*)data;
+   if(!g_network_monitor_get_network_available(dt->monitor) &&
+      g_get_monotonic_time() - dt->begin <= dt->input_seconds * 1000000){
```
cgwalters, line 282: ~1000000 here is `G_USEC_PER_SEC`; it may be simpler actually to do `g_get_monotonic_time() / G_USEC_PER_SEC` instead.~

EDIT: Per above, then we can just do `if (g_get_monotonic_time () > dt->deadline)`.

librepo/handle.c
```diff
@@ -264,6 +266,32 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
     }
 }
 
+struct callback_data {
+   GMainLoop *loop; 
+   long input_seconds;
+   time_t begin;
```
cgwalters, line 272: How about just storing e.g.
```suggestion
    guint64 deadline;
```

which we compute once via:

```
  data_struct.deadline_secs = g_get_monotonic_time() + seconds * G_USEC_PER_SEC;
```

librepo/handle.c
```diff
@@ -264,6 +266,32 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+};
+
+gboolean timeout_callback(gpointer data)
+{ 
+  struct callback_data *dt = (struct callback_data*)data;
+   if(!g_network_monitor_get_network_available(dt->monitor) &&
+      g_get_monotonic_time() - dt->begin <= dt->input_seconds * 1000000){
+      return TRUE;
+   } 
+   g_main_loop_quit(dt->loop);
```
cgwalters, line 285: But wait didn't we lose the `g_network_monitor_can_reach()` bits?  I think we want to do e.g.

```
    if (g_network_monitor_can_reach (dt->monitor, dt->connectable, dt->cancellable, error))  { /* quit loop */ }
```

librepo/handle.c
```diff
@@ -264,6 +266,32 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+   time_t begin;
+   GNetworkMonitor *monitor;
+   GSocketConnectable *connectable; 
+   GCancellable *cancellable;
+};
+
+gboolean timeout_callback(gpointer data)
+{ 
+  struct callback_data *dt = (struct callback_data*)data;
+   if(!g_network_monitor_get_network_available(dt->monitor) &&
```
cgwalters, line 281: One thing I think we should do here is have the availability of the network control whether or not we're polling.  IOW, something like:

```
  if (!g_network_monitor_get_network_available(dt->monitor))
      return FALSE;
```

librepo/handle.c
```diff
@@ -264,6 +266,32 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+      return TRUE;
+   } 
+   g_main_loop_quit(dt->loop);
+   return FALSE;
+}
+
+void on_network_available(GObject *object, GParamSpec *pspec, gpointer data)
+{
+    struct callback_data *dt = (struct callback_data*)data;
+    g_main_loop_run(dt->loop);
```
cgwalters, line 292: Then here instead we do
```
   if (dt->timeout_id == 0 && g_network_monitor_get_network_available (dt->monitor))
     dt->timeout_id = g_timeout_add(200, timeout_callback, &data_struct); 
```

RishabhSaini:
Updated.
I have attempted to resolve the availability of nework control. Kindly review the changes.

librepo/handle.c
```diff
@@ -264,6 +266,50 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+    return G_SOURCE_REMOVE;
+  }
+
+  //if url not reached and time exists, keep source and poll
+  if(!g_network_monitor_can_reach (dt->monitor, dt->connectable, dt->cancellable, NULL) &&
+     g_get_monotonic_time() < dt->deadline_millis){
+    return G_SOURCE_CONTINUE;
+  }
+
+  //if time does not exist, remove source and exit
```
cgwalters, line 299: I am not parsing "if time does not exist".

Would the logic here be a bit cleaner if we did at the very start:

```
  if g_get_monotonic_time() >= dt->deadline_millis) {
    return G_SOURCE_REMOVE;
  }
```

And then here at the very end (along with removing the extra check above) we'd just have something like:

```
    // We're still waiting for the URL to be available, keep polling
    return G_SOURCE_CONTINUE;
```

or so?

librepo/handle.c
```diff
@@ -264,6 +266,50 @@ lr_handle_remote_sources_changed(LrHandle *handle, LrChangedRemoteSource type)
...
+  }
+  
+  //if url reached, remove source and exit loop
+  if(g_network_monitor_can_reach (dt->monitor, dt->connectable, dt->cancellable, NULL)){
+    g_main_loop_quit(dt->loop);
+    return G_SOURCE_REMOVE;
+  }
+
+  //if url not reached and time exists, keep source and poll
+  if(!g_network_monitor_can_reach (dt->monitor, dt->connectable, dt->cancellable, NULL) &&
```
cgwalters, line 294: This condition is redundant

cgwalters (commented):
Thanks!  This is looking pretty good to me.

There's still more to do, but I wouldn't think of these things as strictly blocking:

- The code style/indentation does not match the rest of the project.  librepo seems to use 4 space indents, and spaces between identifier and parenthesis.
- I think to handle cancellation, we actually need to `g_signal_connect (cancellable, "cancelled", ...)` and break out of the mainloop too
- Probably useful to think about what tests could look like for this - unit testing here is inherently hard and needs system integration testing

Now I do not have merge rights on this repository.

librepo/handle.c
```diff
@@ -841,7 +887,62 @@ lr_handle_setopt(LrHandle *handle,
...
+    else if(handle->urls)
+      baseurl = handle->urls[0];
+    assert(baseurl);
+
+    g_autoptr(GUri) uri = g_uri_parse(baseurl, G_URI_FLAGS_NONE, NULL);
+    if(uri == NULL){
+      return FALSE;
+    }
+    const gchar* scheme = g_uri_get_scheme(uri);
+    if(!g_strcmp0(scheme, "file")){
```
cgwalters, line 922: For example, to match the rest of the style this should be
```
    if (!g_strcmp0(scheme, "file")) {
```

Also, be sure you're using 4 spaces and not 2 for indent - across the whole patch.

RishabhSaini:
@m-blaha Friendly Ping: Hey, when do you think this can be merged by

m-blaha (approved):
I'm sorry for delay, the PR somehow dropped off my radar...
Thanks for your work! Just small comment - would it be possible (as a follow-up PR) to add this API also to python bindings?

PR #266, opened 2022-11-21
PGP API extension and fixes

jrohel:
librepo PGP API has been extended:
- support for importing keys from memory and file descriptor
- support for listing/exporting keys (with subkeys)

Added unit tests for new API functions.


jan-kolarik (approved):
Thanks! Nice structured and well documented. Tested locally and checked `valgrind` outputs and everything seems OK.

PR #268, opened 2022-12-01
Update license format to LGPL-2.1-or-later

inknos:
This is NOT relicensing. Due to recent changes in Fedora's License
Format, this commit updates the License field adopting SPDX.

=changelog=
msg: Update License format to LGPL-2.1-or-later
type: enhancement

How to license in Fedora:
https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_field
Packages that need license format updates:
https://pagure.io/copr/license-validate/raw/main/f/packages-without-spdx-final.txt

### DIFF
--- a/librepo.spec
+++ b/librepo.spec
@@ -15,7 +15,7 @@ Version:        1.14.0
 Release:        1%{?dist}
 Summary:        Repodata downloading library
 
-License:        LGPLv2+
+License:        LGPL-2.1-or-later
 URL:            https://github.com/rpm-software-management/librepo
 Source0:        %{url}/archive/%{version}/%{name}-%{version}.tar.gz
 


librepo.spec
```diff
@@ -15,7 +15,7 @@ Version:        1.14.0
 Release:        1%{?dist}
 Summary:        Repodata downloading library
 
-License:        LGPLv2+
+License:        LGPL-2.1-or-later
```
jrohel, line 18: Does "LGPLv2+" really mean "LGPL-2.1-or-later"? Isn't that "LGPL-2.0-or-later"? Then it would mean that PR changes the used license and not just the format.
I'm not saying it's wrong, but at least the commit message is misleading.

inknos: From the fedora command line tool
```
$ license-fedora2spdx LGPLv2+
Warning: more options on how to interpret LGPLv2+. Possible options: ['LGPL-2.0-or-later', 'LGPL-2.1-or-later']
{{pick LGPLv2+ choice}}
```
I guess it is ambiguous.

in https://github.com/rpm-software-management/librepo/blob/master/COPYING says lgpl 2.1 or later. 
I think this clarifies the license field. it is **not** relicensing.

do you want me to add more in the commit message?

jrohel: I quickly read both licenses. 2.1 is a little different.

E.g. paragraph that is extra in version 2.1:
_Use a suitable shared library mechanism for linking with the Library. A suitable mechanism is one that (1) uses at run time a copy of the library already present on the user's computer system, rather than copying library functions into the executable, and (2) will operate properly with a modified version of the library, if the user installs one, as long as the modified version is interface-compatible with the version that the work was made with._

How about writing "Update license to LGPL-2.1-or-later" in the commit message? And the message will be accurate.

inknos: Will do

PR #270, opened 2023-02-02
Bump glib version

fd00:
The GLib URI functions referenced in librepo/handle.c have been added since version 2.66.

- https://docs.gtk.org/glib/struct.Uri.html

Therefore, cmake should check if glib-2.0 is version 2.66 or later.


jrohel:
You're right. Thank you for the fix.

PR #271, opened 2023-02-06
librepo.h: Change how project files are included

mcrha:
Use path-reference in the public header, to avoid ambiguity for generic-like header files from other projects. Otherwise it would depend on the order of the include directories in the build time, which of the header file is picked.

### DIFF
--- a/librepo/CMakeLists.txt
+++ b/librepo/CMakeLists.txt
@@ -42,7 +42,6 @@ SET(librepo_HEADERS
     xmlparser.h
     yum.h
     downloader.h
-    downloader_internal.h
     downloadtarget.h)
 
 ADD_LIBRARY(librepo SHARED ${librepo_SRCS})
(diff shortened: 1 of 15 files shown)


evan-goode:
Thank you, I understand the rationale for this change.

`librepo.h` isn't the only public header; some consumers of librepo #include util.h or yum.h, for instance. It would be best to keep the behavior consistent across all the headers. Are there any drawbacks to using the path-reference style throughout all the headers in the project?

mcrha:
> It would be best to keep the behavior consistent across all the headers.

I agree.

> Are there any drawbacks to using the path-reference style throughout all the headers in the project?

I'm not aware of anything. The public headers should be referenced with the path. I'll update the pull request.

mcrha:
I updated the pull request and split the changes into three commits:

* one changes the `#include`-s to be with the path
* other to not install an internal header
* the last to claim public and internal header files as the project files.

The last change might help the build scripts to properly detect changes in these files, but it's untested.

evan-goode (approved):
Thanks, looks good!

jan-kolarik:
Nice, thanks for the improvement!

PR #272, opened 2023-02-13
Run integration tests for both dnf4 and dnf5

kontura:
Librepo is simultaneously used in both stacks therefore we should test both of them.

This fixes an issue where we currently use dnf5 ci-dnf-stack branch but run dnf4 tests which fails because of missing dependencies (swidtags).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,8 @@ jobs:
         with:
           copr-user: ${{steps.setup-ci.outputs.copr-user}}
 
-  integration-tests:
-    name: DNF Integration Tests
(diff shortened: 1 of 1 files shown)


kontura:
Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/1219

The current CI is failing because it is still running with the old config, with the new configuration it will look like: https://github.com/kontura/librepo/pull/1

jan-kolarik:
Oh I see, I didn't analyze the content of this PR yet and assumed that the requirement from `ci-dnf-stack` is enough :slightly_smiling_face:

jan-kolarik:
LGTM, thanks!

PR #275, opened 2023-05-05
Implement OpenPGP using librpm API

jrohel:
Implements the original librepo OpenPGP API using librpm API instead of GpgMe. It implements its own keyring for public keys. Each key with its subkeys is stored in its own file.
The original GpgMe based implementation was moved to the "gpg_gpgme.c" file. So it is still present and can be activated by the USE_GPGME option in "CMakeList.txt".
This PR leaves the original GpgMe implementation enabled by default in CMakeList.txt. In the .spec file it switches to the librpm API for Fedora >= 39.

Requirement:
A new rpm library is needed that supports OpenPGP ASCII Armored signature parsing. Tested with sequoia based rpm OpenPGP backend.

Missing (requires support in the rpm library):
- Setting the `can_sign` property. It now always returns TRUE.
- Fingerprint for subkeys. An empty string is now returned.
- Return all user IDs. Now only one returns

Notes:
In the Python tests, pgp tests that should succeed were disabled. This is because librepo lacks a Python API for working with OpenGPG keys. The Python tests manipulate the keyring directly using GpgMe. Of course, this only works if the librepo library uses the original GpgMe backend.


CMakeLists.txt
```diff
@@ -32,8 +33,12 @@ PKG_CHECK_MODULES(GLIB2 glib-2.0>=2.66 gio-2.0 REQUIRED)
 PKG_SEARCH_MODULE(LIBCRYPTO REQUIRED libcrypto openssl)
 PKG_CHECK_MODULES(LIBXML2 libxml-2.0 REQUIRED)
 FIND_PACKAGE(CURL 7.52.0 REQUIRED)
-FIND_PACKAGE(Gpgme REQUIRED)
 
+IF (USE_GPGME)
+    FIND_PACKAGE(Gpgme REQUIRED)
+ELSE (USE_GPGME)
+    PKG_CHECK_MODULES(RPM REQUIRED rpm>=4.18.0)
```
j-mracek, line 40: I believe that we should add the conditional require to spec. It might be good to require virtual provide that will signal presence of rpm with right back-end - see https://github.com/rpm-software-management/rpm/issues/2556

librepo/CMakeLists.txt
```diff
@@ -60,9 +67,12 @@ TARGET_LINK_LIBRARIES(librepo
                         ${LIBXML2_LIBRARIES}
                         ${CURL_LIBRARY}
                         ${LIBCRYPTO_LIBRARIES}
-                        ${GPGME_VANILLA_LIBRARIES}
                         ${GLIB2_LIBRARIES}
+                        ${RPM_LIBRARIES}
```
jan-kolarik, line 71: Is there a reason why this is not symmetric as with conditionally adding the `${GPGME_VANILLA_LIBRARIES}` below?

librepo/gpg_rpm.c
```diff
@@ -0,0 +1,613 @@
...
+    g_autofree gchar * bufp = g_new(gchar, size);
+    ssize_t ret = read(fd, bufp, size);
+    if (ret == -1) {
+        const gchar * errstr = g_strerror(errno);
+        g_debug("%s: Error reading from file descriptor %i: %s", __func__, fd, errstr);
+        g_set_error(err, LR_GPG_ERROR, LRE_IO, "Error reading from file descriptor %i: %s", fd, errstr);
+        return -1;
+    }
+    if (ret != size) {
+        g_debug("%s: Detected file size %li but readed %li", __func__, (long)size, (long)ret);
```
jan-kolarik, line 107: Typo: `readed` -> `but %li was read`. Same in the line below...

librepo/gpg_rpm.c
```diff
@@ -0,0 +1,613 @@
...
+        for (int idx = 0; idx < subkeys_count; ++idx) {
+            if (memcmp(pgpDigParamsSignID(subkeys[idx]), keyid, sizeof(pgpKeyID_t)) == 0) {
+                if (dig_params != NULL) {
+                    *dig_params = g_steal_pointer(&subkeys[idx]);
+                }
+                ret = TRUE; // key id match subkey
+                break;
+            }
+        }
+    } while (FALSE);
```
jan-kolarik, line 199: This seems just weird and confusing to me. I understand this has its meaning in macros and such, but why don't just refactor the code here into separate static function?

librepo/gpg_rpm.c
```diff
@@ -0,0 +1,613 @@
...
+        pgpDigParams main_dig_params = NULL;
+        if (pgpPrtParams(key, cert_len, PGPTAG_PUBLIC_KEY, &main_dig_params) == -1) {
+            g_debug("%s: Error: Parsing a OpenPGP packet(s) failed", __func__);
+            g_set_error(err, LR_GPG_ERROR, LRE_GPGERROR, "Parsing a OpenPGP packet(s) failed");
+            return FALSE;
+        }
+        const guint8 * keyid = pgpDigParamsSignID(main_dig_params);
+        const gboolean found = keyring_search_key_id(keyid, NULL, NULL, NULL, home_dir, err);
+        if (found) {
+            g_info("%s: Key with id \"%s\" allready found in keyring", __func__, keyid);
```
jan-kolarik, line 279: typo: `allready` -> `already`

librepo/gpg_rpm.c
```diff
@@ -0,0 +1,613 @@
...
+            g_ptr_array_add(uid_strings, g_strdup(pgpDigParamsUserID(main_dig_params)));
+
+            g_ptr_array_add(uid_strings, NULL);  // add terminating NULL
+            lr_key.uids = (char **)g_ptr_array_free(uid_strings, FALSE);
+
+            lr_key.subkeys = (LrGpgSubkey *)(gsubkeys->len > 0 ? g_array_free(gsubkeys, FALSE) : g_array_free(gsubkeys, TRUE));
+            lr_key.raw_key = export_keys ? pgpArmorWrap(PGPARMOR_PUBKEY, (const unsigned char *)buf, len) : NULL;
+            g_array_append_val(keys, lr_key);
+
+        } while (FALSE);
```
jan-kolarik, line 464: Again, I think it would be better to refactor this.

kontura:
I think the CI should be fixed by a rebase.

librepo/gpg_rpm.c
jrohel: The single-pass loop here allows `break` to be used instead of `goto`.
It could be solved in another way. For example, with more `if`s. But that would be rather more complicated and harder to read. Or replace `break` with `return`. But then I would have to ensure that the resources are freed before each `return`.

jrohel: Thanks.

librepo/gpg_rpm.c
jrohel: Thanks

librepo/gpg_rpm.c
jrohel: Thanks

librepo/CMakeLists.txt
jrohel: No. Forgotten from testing.

CMakeLists.txt
jrohel: Thanks. I modified .spec file.

librepo/gpg_rpm.c
jan-kolarik: OK, I am not that familiar with conventions in C, so if it's commonly used in such cases, let's keep it as it is.

jan-kolarik (approved):
LGTM

Issue #276, opened 2023-05-22
The `lr gpg list_keys` function crashes if called on an empty keyring


jrohel:
Error in line:
`for (guint i = 0; i < keys->len - 1; ++i) {`

`keys->len` is unsigned. For empty keyring `keys->len == 0`.
`0 - 1` is max value for unsigned type.

xiaoge1001:
@jrohel Hi, has this problem been solved? Is there any reproduction code?

jrohel:
@xiaoge1001
> has this problem been solved?

Yes, in this PR https://github.com/rpm-software-management/librepo/pull/277

> Is there any reproduction code?

To reproduce the error, you need to call the `lr gpg list_keys` librepo API function where the `home_dir` parameter points to an empty directory. Or a directory containing an empty keyring.

The `lr gpg list_keys` function is used by the `libdnf5` library and therefore the `dnf5` application can be used to reproduce the error.
The older library `libdnf` (and thus the application dnf, microdnf, PackageKit, ...) does not use this function yet, it assumes that librepo uses GpgME and accesses the keyring directly via the GpgME library. So they cannot be used for reproduction.

A simple reproducer:
- use a repository with `repo_gpgcheck=1` (e.g. google-chrome)
- run dnf5 and import the keys
- delete the contents of the directory with the imported keys (the path to the imported keys for the root user `/var/cache/libdnf5/<repo_cache_dir>/pubring/`)
- run dnf5 again, this will call the `lr gpg list_keys` function on an empty directory (contains 0 keys)

jrohel:
I am closing the issue. It is fixed by PR https://github.com/rpm-software-management/librepo/pull/277

PR #277, opened 2023-05-22
PGP: Fix lr_gpg_list_keys: Add "key->len" value test

jrohel:
Solves: https://github.com/rpm-software-management/librepo/issues/276


librepo/gpg.c
```diff
@@ -393,7 +393,7 @@ lr_gpg_list_keys(gboolean export_keys, const char *home_dir, GError **err)
             subkey = subkey->next;
         }
          // All subkeys in the list except the last one are followed by another subkey
-        if (subkeys->len > 0) {
+        if (subkeys->len > 1) {
```
j-mracek, line 396: May I ask you why the first subkey is skipped? It was skipped even by original code. The description says something about the last key, but it skips the first one. May I ask you to enhance the comment?

jrohel: The first key is not skipped. But if there is only one key in the list, the first key is also the last key. It is unnecessary to continue with the following cycle. It would run zero times anyway. So changing `keys->len > 0` to `keys->len > 1` is just an optimization.

j-mracek: May I ask you to enhance comment in lines 395 and 422. I would like to see that the code bellow mark keys that are followed by ` // Mark all subkeys in the list except the last one that they are followed by another subkey`

jrohel: OK. I modified the comment.

PR #278, opened 2023-06-12
We cannot use the same copr build for dnf4 and dnf5 CI runs

kontura:
In dnf4 copr build we need microdnf but it is obsoleted by dnf5 (in F38) therefore we need separate builds.

This is an alternative to: https://github.com/rpm-software-management/ci-dnf-stack/pull/1286

Before this we have to fix the microdnf build though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,8 @@ name: DNF CI
 on: pull_request_target
 
 jobs:
-  copr-build:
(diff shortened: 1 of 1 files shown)


kontura:
The failing CI is expected because the workflow is taken from master.
You can check my example PR: https://github.com/kontura/librepo/pull/2 where the commit is already in master.

jan-kolarik:
Thanks!

Issue #281, opened 2023-09-12
If librepo is built with -DUSE_GPGME and RPM with -DWITH_INTERNAL_OPENPGP=ON -DWITH_OPENSSL=ON, test_gpg_check_signature fails: Assertion 'ret' failed


kloczek:
cmake setup
```console
-- Cache values
CMAKE_BUILD_TYPE:STRING=RelWithDebInfo
CMAKE_INSTALL_PREFIX:PATH=/usr
ENABLE_DOCS:BOOL=ON
ENABLE_PYTHON:BOOL=ON
ENABLE_TESTS:BOOL=ON
USE_GPGME:BOOL=OFF
WITH_ZCHUNK:BOOL=ON
```
and test suite is failing with:
```console
+ cd librepo-1.16.0
+ /usr/bin/ctest --test-dir x86_64-redhat-linux-gnu --output-on-failure --force-new-ctest-process -j48
Internal ctest changing into directory: /home/tkloczko/rpmbuild/BUILD/librepo-1.16.0/x86_64-redhat-linux-gnu
Test project /home/tkloczko/rpmbuild/BUILD/librepo-1.16.0/x86_64-redhat-linux-gnu
    Start 1: test_main
    Start 2: test_python
1/2 Test #1: test_main ........................***Failed    0.12 sec
Tests using directory: /tmp/librepoWXh22S
Running suite(s): checksum
 gpg
 handle

(process:685827): librepo-WARNING **: 02:07:36.158: WARNING! Deprecated LRO_MIRRORLIST used

(process:685827): librepo-WARNING **: 02:07:36.158: WARNING! Deprecated LRO_MIRRORLIST used
 internal_mirrorlist
 metalink
 mirrorlist
 package_downloader
 repoconf
 repomd
 url_substitution
 util
 version
95%: Checks: 47, Failures: 2, Errors: 0
/home/tkloczko/rpmbuild/BUILD/librepo-1.16.0/tests/test_gpg.c:51:F:Main:test_gpg_check_signature:0: Assertion 'ret' failed
/home/tkloczko/rpmbuild/BUILD/librepo-1.16.0/tests/test_gpg.c:169:F:Main:test_gpg_check_key_export:0: Assertion 'subkeys != NULL' failed: subkeys == 0

2/2 Test #2: test_python ......................   Passed   12.24 sec

50% tests passed, 1 tests failed out of 2

Total Test time (real) =  12.25 sec

The following tests FAILED:
[...]

PR #282, opened 2023-09-15
lr_gpg_check_signature: Forward PGP error messages from RPM

ppisar:
When debugging a test failure with RPM using internal OpenPGP+OpenSSL implementation (that's a bug in RPM, not in librepo), I discovered that librepo tests did not print error messages and that lr_gpg_check_signature() did not forwarded an error message from RPM.

RPM before rpm-4.19.0-alpha2 did not provided provided any error messages. That has changed with new functions pgpPrtParams2() and pgpVerifySignature2().

This patch enhances librepo code to use the new RPM functions if available and to propagate the RPM error messages via an already existing GError argument.

This patch also enhances librepo tests to actually print the unexpected error messages.

Both enhancements should help people to debug their failures.

Nonetheless, internal OpenPGP implementation in RPM does not set any error messages and that will probably not change because RPM is going to remove that implementation. On the other hand, Sequoia implementation in RPM forwards the messages from Sequoia library. Yet I was unbable to obtain any message. Sequoia promissed to improve their error messaging, especially with a demise of SHA-1. So I believe this librepo enhancement is useful.

Implementation details: I wrapped pgpPrtParams2() into a function because it's called at multiple places. Contrary I did not wrap pgpVerifySignature2() because it's called only at one place.

https://github.com/rpm-software-management/librepo/issues/281


jrohel:
Thank you. LGTM

PR #285, opened 2023-09-29
PGP: Enable creating a UID directory for GnuGP agent socket in /run/gnupg/user

ppisar:
librepo used to precreate the directory in /run/user to make sure a GnuPG agent executed by GPGME library, one of librepo's PGP backends, places its socket there.

The UID directories there are normally created and removed by systemd (logind PAM session). librepo created them for a case when a package manager is invoked out of systemd session, before the super user logs in. E.g. by a timer job to cache repository metadata.

A problem was when this out-of-session process was a SELinux-confined process creating files with its own SELinux label. Then the UID directory was created with a SELinux label different from the one expected by systemd.

Later, the mismatching label interfered with systemd when logging out a corresponding user.

This patch fixes it by creating the UID directories in a different path which is not managed by systemd. That path is /run/gnupg/user. GnuPG recognizes the path if it is compiled with
--enable-run-gnupg-user-socket option.

Unfortunatelly, this patch cannot keep precreating directories in the old /run/user path for compatibility with GnuPG built without that option because that would still break systemd. Also GPGME does not provide a way for checking how GnuPG was configured. Therefore users who rely on precreating the directories by librepo need to keep their librepo and GnuPG in synchronization. So this patch adds a new -DUSE_RUN_GNUPG_USER_SOCKET=ON CMake option to use /run/gnupg/user. A default behavior is using the old /run/user.

https://issues.redhat.com/browse/RHEL-10720


jrohel:
LGTM

PR #286, opened 2023-10-05
PGP: fix: Support importing binary public keys in librpm backend, add unit tests

jrohel:
The `lr_gpg_import_key*` functions in the GpgMe backend support the import of binary public keys.
The patch adds support for importing binary public keys to a backend that uses the librpm API. Thus eliminating the regression.

The PR also adds unit tests for importing binary public keys.


jan-kolarik (approved):

jan-kolarik:
Thanks!

jan-kolarik:
> @jan-kolarik Can we get this backported to Fedora and into F39 before the release?

Oh, sure. I was checking the upstream components more than a week ago last time, and I see that particularly for librepo, it's the right time to make a new release. Thanks for the bump!

PR #289, opened 2023-10-18
PGP: Set a default creation SELinux labels on GnuPG directories

ppisar:
This is another way how to fix mismatching SELinux context on /run/user directories without moving the directories to /run/gnupg/user.

librepo used to precreate the directory in /run/user to make sure a GnuPG agent executed by GPGME library places its socket there.

The directories there are normally created and removed by systemd (logind PAM session). librepo created them for a case when a package manager is invoked out of systemd session, before the super user logs in. E.g. by a timer job to cache repository metadata.

A problem was when this out-of-session process was a SELinux-confined process creating files with its own SELinux label different from a DNF program. Then the directory was created with a SELinux label different from the one expected by systemd and when logging out a corresponding user, the mismatching label clashed with systemd.

This patch fixes the issue by choosing a SELinux label of those directories to the label defined in a default SELinux file context database.

This patch adds a new -DENABLE_SELINUX=OFF CMake option to disable the new dependency on libselinux. A default behavior is to support SELinux only if GPGME backend is selected with -DUSE_GPGME=ON.

https://issues.redhat.com/browse/RHEL-10720


ppisar:
The CI test fails because of misconfigured CI (rhel-8.10 branch is built and tested in Fedora 38 environment and then installation of built librepo RPM pakcages fails because a system has installed a far higher librepo NEVRA).

jan-kolarik (approved):

PR #294, opened 2024-02-09
Avoid libc buffered IO

stewartsmith:
the FILE related IO functions in libc do buffering inside the C library, and are generally less performant than using the file descriptor based open()/read()/write() functions.

The FILE based approach somewhat limits the maximum throughput of librepo, as well as increases CPU usage. In my benchmarks of a reposync of the Amazon Linux 2023 x86-64 repositories, this move to file descriptor based IO saves about 1 second of user time, and .5 seconds of system time, for a wall clock time benefit of a few seconds (102s vs 99s).

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -148,8 +148,8 @@ typedef struct {
         Current protocol */
     CURL *curl_handle; /*!<
         Used curl handle or NULL */
-    FILE *f; /*!<
-        fdopened file descriptor from LrDownloadTarget and used
+    int fd; /*!<
+        opened file descriptor from LrDownloadTarget and used
         in curl_handle. */
     char errorbuffer[CURL_ERROR_SIZE]; /*!<
         Error buffer used in curl handle */
@@ -283,7 +283,7 @@ typedef struct {
  *       | LrDownloadTarget *target  -----------/   | int fd                   |
  *       | LrMirror *mirror          --------/      | LrChecksumType checks..  |
  *       | CURL *curl_handle          |-+           | char *checksum           |
- *       | FILE *f                    |             | int resume               |
+ *       | int fd                     |             | int resume               |
  *       | GSList *tried_mirrors      |             | LrProgressCb progresscb  |
  *       | gint64 original_offset     |             | void *cbdata             |
  *       | GSlist *lrmirrors         ---\           | GStringChunk *chunk      |
@@ -614,22 +614,51 @@ lr_zck_writecb(char *ptr, size_t size, size_t nmemb, void *userdata)
 size_t
 lr_writecb(char *ptr, size_t size, size_t nmemb, void *userdata)
 {
-    size_t cur_written_expected = nmemb;
-    size_t cur_written;
     LrTarget *target = (LrTarget *) userdata;
     #ifdef WITH_ZCHUNK
     if(target->target->is_zchunk && !target->range_fail && target->mirror->mirror->protocol == LR_PROTOCOL_HTTP)
         return lr_zck_writecb(ptr, size, nmemb, userdata);
     #endif /* WITH_ZCHUNK */
 
-    gint64 all = size * nmemb;  // Total number of bytes from curl
     gint64 range_start = target->target->byterangestart;
     gint64 range_end = target->target->byterangeend;
 
+    /* libcurl docs tell us size is always 1, but don't bet on it */
+    size_t all;
+    ssize_t r;
+    ssize_t written;
+    /*
+     * We write up to 32MB at a time, mainly to ensure that the below loops
+     * never bitrot, and it's regularly tested with real world RPMs.
+     */
+    #define WRITECB_CHUNK_MAX 32*1024*1024
+    /*
+     * The libcurl docs point to size always being 1, and the max value for
+     * nmemb being CURL_MAX_WRITE_SIZE (default 16kb, compile time setting),
+     * and if CURLOPT_HEADER is set, then it's CURL_MAX_HTTP_HEADER, usually
+     * meaning 100k.
+     */
+    all = size * nmemb;
+    written = 0;
+    if ((size != 0 && all / size != nmemb) || all > SSIZE_MAX) {
+        /*
+	 * But if libcurl changes (likely breaking the world), let's handle the
+	 * overflow by just writing one chunk at a time until we're done.
+	 */
+        all = WRITECB_CHUNK_MAX;
+    }
+
     if (range_start <= 0 && range_end <= 0) {
-        // Write everything curl give to you
-        target->writecb_recieved += all;
-        return fwrite(ptr, size, nmemb, target->f);
+        while(( all - written ) > 0) {
+	    r = write(target->fd, ptr + written, ((all - written) > WRITECB_CHUNK_MAX)? WRITECB_CHUNK_MAX : all - written);
+	    if (r < 0 && errno != EINTR) {
+	        g_warning("Error while writing file: %s", g_strerror(errno));
+		return (size_t)written;
+	    }
+	    written += r;
+	    target->writecb_recieved += r;
+	}
+        return (size_t)written;
     }
 
     /* Deal with situation when user wants only specific byte range of the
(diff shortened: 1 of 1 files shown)


stewartsmith:
For reference, my benchmarking has been done on a `m5n.16xlarge` EC2 instance to the in-region S3 buckets as well as to the CDN repositories. That instance type has 256GB memory, a 75Gbit network connection, and is a 64 core Cascade Lake system.

The background of this is that a *lot* of EC2 instances don't live that long (relatively speaking), and never install RPMs except on launch - so all the time-to-install RPMs is time spent scaling up a system that could be better served by running the customer workload.

librepo/downloader.c
```diff
@@ -619,7 +619,7 @@ lr_writecb(char *ptr, size_t size, size_t nmemb, void *userdata)
     if (range_start <= 0 && range_end <= 0) {
         // Write everything curl give to you
         target->writecb_recieved += all;
-        return fwrite(ptr, size, nmemb, target->f);
+        return write(target->fd, ptr, all);
```
ppisar, line 622: Replacing fwrite() with write() is not so simple:

* write() expects size_t in its last argument, while you pass gint64. That won't match on 32-bit memory model, e.g. on 32-bit x86 platform.
* write() returns a signed ssize_t, while fwrite() unsigned size_t. Your return statement does match lr_writecb() prototype.
* write() can return before writing all data with errno==EINTR and you need to call write() again.

Please fix these issues at all places you replaced fwrite() with write().

librepo/downloader.c
```diff
@@ -684,7 +684,7 @@ lr_writecb(char *ptr, size_t size, size_t nmemb, void *userdata)
     }
 
     assert(nmemb > 0);
-    cur_written = fwrite(ptr, size, nmemb, target->f);
+    cur_written = write(target->fd, ptr, size * nmemb);
```
ppisar, line 687: What if size*nmemb overflows size_t? fwrite() handled it for you. Now you need to implement it check yourself.

librepo/downloader.c
```diff
@@ -1033,9 +1033,9 @@ remove_librepo_xattr(LrDownloadTarget * target)
 gboolean
 lr_zck_clear_header(LrTarget *target, GError **err)
 {
-    assert(target && target->f && target->target && target->target->path);
+    assert(target && target->fd && target->target && target->target->path);
```
ppisar, line 1036: fd is an int. fd==0 is a valid descriptor. You should compare it to -1, or at least to < 0 as in invalid descriptor.

librepo/downloader.c
```diff
@@ -1295,7 +1295,7 @@ static gboolean
 check_zck(LrTarget *target, GError **err)
 {
     assert(!err || *err == NULL);
-    assert(target && target->f && target->target);
+    assert(target && target->fd && target->target);
```
ppisar, line 1298: Again. fd == 0 is a valid value. Compare to -1.

librepo/downloader.c
```diff
@@ -1538,15 +1528,15 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
             target->curl_handle = NULL;
             g_free(target->headercb_interrupt_reason);
             target->headercb_interrupt_reason = NULL;
-            fclose(target->f);
-            target->f = NULL;
+            close(target->fd);
+            target->fd = 0;
```
ppisar, line 1532: Assign -1.

librepo/downloader.c
```diff
@@ -1689,9 +1678,9 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
         curl_easy_cleanup(target->curl_handle);
         target->curl_handle = NULL;
     }
-    if (target->f != NULL) {
-        fclose(target->f);
-        target->f = NULL;
+    if (target->fd != 0) {
```
ppisar, line 1681: Compare to -1 and set to  -1.

librepo/downloader.c
```diff
@@ -2353,8 +2341,8 @@ check_transfer_statuses(LrDownload *dd, GError **err)
         target->curl_handle = NULL;
         g_free(target->headercb_interrupt_reason);
         target->headercb_interrupt_reason = NULL;
-        fclose(target->f);
-        target->f = NULL;
+        close(target->fd);
+        target->fd = 0;
```
ppisar, line 2345: Set to -1.

librepo/downloader.c
```diff
@@ -2756,8 +2744,8 @@ lr_download(GSList *targets,
             curl_multi_remove_handle(dd.multi_handle, target->curl_handle);
             curl_easy_cleanup(target->curl_handle);
             target->curl_handle = NULL;
-            fclose(target->f);
-            target->f = NULL;
+            close(target->fd);
+            target->fd = 0;
```
ppisar, line 2748: Set to -1.

librepo/downloader.c
```diff
@@ -2803,7 +2791,7 @@ lr_download(GSList *targets,
     for (GSList *elem = dd.targets; elem; elem = g_slist_next(elem)) {
         LrTarget *target = elem->data;
         assert(target->curl_handle == NULL);
-        assert(target->f == NULL);
+        assert(target->fd == 0);
```
ppisar, line 2794: Compare against -1.

ppisar:
In general, I like your patch, but please address the different semantics I pointed in-line.

stewartsmith:
Thanks for the eyes on it. I can clean up the review comments, and spin a rev2.

librepo/downloader.c
```diff
@@ -1573,8 +1563,7 @@ prepare_next_transfer(LrDownload *dd, gboolean *candidatefound, GError **err)
 
         if (target->original_offset == -1) {
             // Determine offset
-            fseek(target->f, 0L, SEEK_END);
-            gint64 determined_offset = ftell(target->f);
+            gint64 determined_offset = lseek(target->fd, 0L, SEEK_END);
```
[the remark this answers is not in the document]
stewartsmith: fixed.

there's this snipped below from the original code:
```
        gint64 used_offset = target->original_offset;
        g_debug("%s: Used offset for download resume: %"G_GINT64_FORMAT,
                __func__, used_offset);

        c_rc = curl_easy_setopt(h, CURLOPT_RESUME_FROM_LARGE,
                                (curl_off_t) used_offset);
```

So there might be some places in existing code that doesn't match everywhere. Possible that this should all just move to `off_t` ?

librepo/downloader.c
stewartsmith: fixed.

Included some code for keeping to reasonable IO sizes and just avoiding any of these issues with signed/unsigned.

librepo/downloader.c
stewartsmith: Updated to check for overflow and fall back to a well known size.

librepo/downloader.c
stewartsmith: fixed

librepo/downloader.c
stewartsmith: fixed

librepo/downloader.c
stewartsmith: fixed

librepo/downloader.c
stewartsmith: fixed

librepo/downloader.c
stewartsmith: fixed.

librepo/downloader.c
stewartsmith: fixed.

librepo/downloader.c
stewartsmith: fixed

stewartsmith:
I think I've managed to address the comments and ensure correct behavior in (hopefully) all error conditions.

Conan-Kudo (changes requested):
Overall, it looks fine. One small change, though.

librepo/downloader.c
```diff
@@ -604,22 +604,51 @@ lr_zck_writecb(char *ptr, size_t size, size_t nmemb, void *userdata)
...
 
+    /* libcurl docs tell us size is always 1, but don't bet on it */
+    size_t all;
+    ssize_t r;
+    ssize_t written;
+    /*
+     * We write up to 32MB at a time, mainly to ensure that the below loops
+     * never bitrot, and it's regularly tested with real world RPMs.
+     */
+#define WRITECB_CHUNK_MAX 32*1024*1024
```
Conan-Kudo, line 624: Please indent to match the line starting point.

Conan-Kudo (changes requested):

stewartsmith:
I'm going to close this PR as 1) it's pretty old and 2) with my other improvements, I can't currently benchmark this change as being meaningful (yet).

Issue #298, opened 2024-02-12
WITH_ZCHUNK macro affects ABI


ppisar:
DNF5 was asked to drop a useless dependency on zchunk <https://github.com/rpm-software-management/dnf5/issues/1233> because it delegates all zchunk work to librepo. However it turned out that `struct LrDownloadTarget` ABI defined in `<librepo/downloadtarget.h>` changes with `WITH_ZCHUNK` macro.

More over, if the macro is defined, `<zck.h>` is included albeit `librepo-devel` RPM package does not depend on `zchunk-devel`.

All that pushes librepo internals to it users, e.g. to DNF5, and makes the build chain more error prone because the whole aplication stack needs to keep WITH_ZCHUNK in synchronization. 

I would appreciate isolating librepo ABI from WITH_ZCHUNK macro.

PR #299, opened 2024-02-12
Change header files to match a configured ABI regarding a zchunk support

ppisar:
If librepo was built with -DWITH_ZCHUNK=ON but an application included <librepo/librepo.h> without defining -DWITH_ZCHUNK, struct LrDownloadTarget mismatched. The same problem happened vice versa. Following the correct ABI was a burden for librepo applications and a mistake there could lead to a memory corruption.

This patch replaces the WITH_ZCHUNK macros in the public header files with a constant so that the installed header files do not depend on the macro and always match the installed library.

This patch also adds a missing RPM dependency on zchunk-devel to librepo-devel if built with enabled zchunk. (Some header files include <zck.h>.)

An implementation note: CMakeLists.txt appends the header search paths with both a top-level build directory as well as with a librepo subdirectory. The latter is for "util.h" inclusions from the C files, the former is for <librepo/util.h> inclusions from other librepo header files. Without the former a C preprocesor would use already installed <librepo/util.h>. We do not want to infect the being built library with old and possible nonmatching system header files.

Fixes: #298

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,9 +70,10 @@ IF (USE_RUN_GNUPG_USER_SOCKET)
 ENDIF (USE_RUN_GNUPG_USER_SOCKET)
 
 IF (WITH_ZCHUNK)
-PKG_CHECK_MODULES(ZCHUNKLIB zck>=0.9.11 REQUIRED)
-SET (CMAKE_C_FLAGS          "${CMAKE_C_FLAGS} -DWITH_ZCHUNK")
-SET (CMAKE_C_FLAGS_DEBUG    "${CMAKE_C_FLAGS_DEBUG} -DWITH_ZCHUNK")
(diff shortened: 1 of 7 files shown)


kontura:
Thank you!

The failing tests are not related.

ppisar:
After merging this pull request,  "dnf/zchunk.feature:141  using mirror wihtout ranges supports and zchunk results in only two GET requests for primary (the first try is with range specified)" scenario from ci-dnf-stack started to fail. I'm investigating it.

ppisar:
> After merging this pull request, "dnf/zchunk.feature:141 using mirror wihtout ranges supports and zchunk results in only two GET requests for primary (the first try is with range specified)" scenario from ci-dnf-stack started to fail. I'm investigating it.

A fix proposed  in PR #302.

PR #301, opened 2024-03-06
Fix some compiler warnings

ppisar:
This patch set resolves these compiler warnings:

* enumerator value for ‘LR_AUTH_ONLY’ is not an integer constant expression
* format ‘%p’ expects argument of type ‘void *’
* initialization discards ‘const’ qualifier from pointer target type
* _BSD_SOURCE and _SVID_SOURCE are deprecated

Warnings regarding new python and new curl will be addressed in separate pull requests as they will probably need to implement fallback code for older libraries.


kontura:
The failing tests are not related.

PR #302, opened 2024-03-08
Fix building zchunk code if zchunk is enabled

ppisar:
Commit 66c99da1206c96fefc1b8279f777afefb79dc614 (Change header files to match a configured ABI regarding a zchunk support) accidentally removed a definition of WITH_ZCHUNK C preprocessor symbol. That symbol controls including zchunk support into the library's code.

As a result, librepo library was always missing a zchunk support and ci-dnf-stack "dnf/zchunk.feature:141 using mirror wihtout ranges supports and zchunk results in only two GET requests for primary (the first try is with range specified)" always failed.

This fix returns back the macro into CFLAGS.


ppisar:
@kontura, you reviewed a pull request which introduced a regression. Could you review this pull request which fixes the regression?

The desired output is that when building librepo (as configured in ci-dnf-stack), a compiler should have -DWITH_ZCHUNK option. I suspect that this fix will have an effect on ci-dfn-stack after building nightlies with the fix. Not immediately when merging this pull request. At least that's what we observed on the breaking pull request.

kontura:
It is odd that the specific CI test passed on the earlier PR.

kontura:
> The desired output is that when building librepo (as configured in ci-dnf-stack), a compiler should have -DWITH_ZCHUNK option. I suspect that this fix will have an effect on ci-dfn-stack after building nightlies with the fix. Not immediately when merging this pull request. At least that's what we observed on the breaking pull request.

That would suggest we are not actually using the build from the PR for the CI run.
I will look into it.

ppisar:
I looked at here failed "DNF4 Integration Tests". It updates librepo from copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly repository (1.17.0-20240308004650.6.g66c99da.fc39) instead of upgrading to the one build in "DNF4 Copr Build" task.

ppisar:
>  the one build in "DNF4 Copr Build" task.

That one is 1.17.0-20240308125436.7.gcdc88a7.fc39.

Issue #305, opened 2024-04-05
If zchunk is supported, `librepo` pkgconfig is missing a dependency on zchunk


kloczek:
With 66c99da1 has been changed ABI and when librepo is build with zchunk  support `<librepo/downloadtarget.h>` header needs `<zck.h>`.
Issue is that if librepo is build that way content of the installed librepo.pc has no `zchunk`  in `Requires:` field.
This as result causes that for example PackageKit build fails with
```console
FAILED: backends/dnf/libpk_backend_dnf.so.p/dnf-backend.c.o
/usr/bin/gcc -Ibackends/dnf/libpk_backend_dnf.so.p -Ibackends/dnf -I../backends/dnf -Isrc -I../src -Ilib/packagekit-glib2 -I../lib/packagekit-glib2 -Ilib -I../lib -I. -I.. -I/usr/include/gl
ib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/gio-unix-2.0 -I/usr/include/appstream -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=c99 -DHAVE_POLKIT_0_114=1 -DHAVE_SYSTEM
D_SD_DAEMON_H=1 -DHAVE_SYSTEMD_SD_LOGIN_H=1 -DPK_BUILD_LOCAL=1 -DPK_ENABLE_DAEMON_TESTS=1 -DPLYMOUTH_0_9_5=1 -D_DEFAULT_SOURCE -DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_44 -DPK_COMPILATION
 -Wall -Wcast-align -Wno-uninitialized -Wmissing-declarations -Wredundant-decls -Wpointer-arith -Wcast-align -Wwrite-strings -Winit-self -Wreturn-type -Wformat-nonliteral -Wformat-security
-Wmissing-include-dirs -Wmissing-format-attribute -Wclobbered -Wempty-body -Wignored-qualifiers -Wsign-compare -Wtype-limits -Wuninitialized -Wno-unused-parameter -Waggregate-return -Wdecla
ration-after-statement -Wshadow -Wno-strict-aliasing -Winline -Wmissing-parameter-type -Woverride-init -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protect
[...]

PR #307, opened 2024-04-05
Add a private dependency on zck to librepo.pc if zchunk support is enabled

ppisar:
Add a private dependency on zck to librepo.pc if zchunk support is enabled
    
If librepo is built with zchunk support, zchunk's zck.h header file is
included from librepo public header files, thus applications building
against librepo needs zchunk header files available at build time.
Those who only use pkg-config for tracking dependencies, reported
build failures:
    
        /usr/include/librepo/downloadtarget.h:27:10: fatal error: zck.h: No such file or directory
    
This patch adds the dependency on zck to librepo.pc.
    
It enhances Requires.private instead of Requires to prevent from
contaminating "pkgconf --libs librepo" with -lzck. Contrary, "pkgconf
--cflags librepo" correctly appends include path to zck header files.
The dependency on zck.pc is only checked and enforced with "pkgconf
--cflags librepo" invocation. This behaviour is a feature. See
<https://github.com/pkgconf/pkgconf/issues/352> and
<https://github.com/pkgconf/pkgconf/issues/300>.
    
Fixes: #305

PR #308, opened 2024-04-05
RFC: Add zchunk header path to librepo.pc if zchunk support is enabled

ppisar:
This patch is a request for comments. It improves https://github.com/rpm-software-management/librepo/issues/305, but does not do a naive fix by adding a pkg-config dependency because of overlinking. Cf. #307 with overlinking.

If librepo is built with zchunk support, zchunk's zck.h header file is included from librepo public header files, thus applications building against librepo needs zchunk header files available at build time. Those who only use pkg-config for tracking dependencies, reported build failures:

    /usr/include/librepo/downloadtarget.h:27:10: fatal error: zck.h: No such file or directory

This patch does not adds the dependency on zck to librepo.pc. Instead it adds zchunk header path the librepo.pc and documenta the dependency in README.md.

Reason for not adding the explicit dependency is that "pkg-config --libs lirepo" would return unnecessay -lzck and lead to an overlinking. There is no non-library kind of dependency in pkg-config. (Ideal solution would be if pkg-config supported those dependencies, or if zck provided a non-library pkg-config file in addition to zck.pc.)

Fixes: #305


ppisar:
I found a better fix, implemented in #307.

PR #309, opened 2024-04-25
Hash cache: Improved work with extended file attributes

jrohel:
Do not work with xattrs after finding out that they are not supported.


m-blaha (approved):
Thanks!

PR #312, opened 2024-06-04
Fix error handling, Fix examples and build them

jrohel:
- Fix: Possible usage of uninitialized value in lr_handle_network_wait

- Fix: `lr_gpg_import_key`, `lr_gpg_import_key_from_fd`: Don't ignore input file errors

- Fix: Removal of warnings "don't set value that will not be read"

- The "examples" directory is part of the librero. The "download_low_level_api.c" example contained an error that prevented it from compiling. The examples were not compiled and therefore the error was ignored for a long time. Fixed the C examples and added CMakeFile.txt to build them.


jan-kolarik (approved):
Thanks!

PR #313, opened 2024-06-04
gpg: Check is_selinux_enabled() before trying to label

cgwalters:
The default for container execution is that `/sys/fs/selinux` is not mounted, and the libselinux library function `is_selinux_enabled` should be used to dynamically check if the system should attempt to perform SELinux labeling.

This is how it's done by rpm, ostree, and systemd for example.

But this code unconditionally tries to label if it finds a policy, which breaks in an obscure corner case
when executed inside a container when we're not using overlayfs for the backend.


cgwalters:
Corresponding libdnf PR https://github.com/rpm-software-management/libdnf/pull/1665

ppisar (approved):
Good.

PR #314, opened 2024-06-04
Improve performance of large number of package downloads

stewartsmith:
For a reposync of Fedora 40, this reduces the loop in lr_download_packages() from taking ~61 seconds to ~1second.


cgwalters (approved):
Ah, classic.  (xref https://rust-unofficial.github.io/too-many-lists/#an-obligatory-public-service-announcement )

It looks like the linked list usage is embedded in public APIs here.  Oh well.

PR #318, opened 2024-07-08
Use rpm-sequoia on RHEL 10+

yselkowitz:
RHEL 10 also has a sufficiently new RPM with rpm-sequoia enabled.

This avoids test failures in ELN, e.g. https://koji.fedoraproject.org/koji/taskinfo?taskID=120187681 with gpgme vs https://koji.fedoraproject.org/koji/taskinfo?taskID=120191938 with rpm-sequoia (this patch).

### DIFF
--- a/librepo.spec
+++ b/librepo.spec
@@ -8,7 +8,7 @@
 %bcond_without zchunk
 %endif
 
-%if 0%{?fedora} >= 39
+%if 0%{?fedora} >= 39 || 0%{?rhel} >= 10
 %bcond_with use_gpgme
 %bcond_with use_selinux
 %else


yselkowitz:
ping @jan-kolarik @evan-goode 

1.18.0 is FTBFS on ELN which is breaking our composes.  Downstream patch is ready too: https://src.fedoraproject.org/rpms/librepo/pull-request/6

ppisar:
Do you have an idea what has changed that the tests started to segfault? I was unable to find the cause as I was unable to reproduce it locally or in a local mock. Also in RHEL Koji it suddenly stopped crashing yesterday.

ppisar (approved):
It makes sense. RPM uses sequoia in RHEL 10 since 2022-11-24 (002373bffdb75af81bf2bdef6c0b06b6d5616922 commit).

PR #319, opened 2024-07-11
spec: Correct setting -DENABLE_SELINUX cmake argument

ppisar:
%{?need_selinux:ON} always expanded to "ON" because %{?} tests whether a macro is defined. Not whether it is true.

The fixed problem is so far only cosmetic becasue CMakeLists.txt ignores ENABLE_SELINUX if USE_GPGME cannot be enabled.


ppisar:
Python tests also fail because of it. I will amend them.

ppisar:
I fixed also the Python test.

ppisar:
Ignore the last two comments. They belonged to #320.

m-blaha (approved):
Thank you!

PR #320, opened 2024-07-11
tests: Adapt to CURL without NTLM support

ppisar:
If CURL is built without NTLM support (e.g. if libcurl-minimal RPM package is installed instead of libcurl on Fedora), tests/test_handle.c failed:

    /home/test/librepo/tests/test_handle.c:61:F:Main:test_handle:0: Assertion 'lr_handle_setopt(h, ((void *)0), LRO_HTTPAUTHMETHODS, LR_AUTH_NTLM)' failed

The cause is that the test exhibing NTLM authentication also checks that lr_handle_setopt() succeeds.

This patch stops checking a return value of lr_handle_setopt() in case of LR_AUTH_NTLM because a meaning of the test is checking for memory leaks.

### DIFF
--- a/tests/python/tests/test_handle.py
+++ b/tests/python/tests/test_handle.py
@@ -474,14 +474,22 @@ class TestCaseHandle(unittest.TestCase):
         self.assertEqual(h.httpheader, None)
 
         self.assertEqual(h.httpauthmethods, librepo.LR_AUTH_BASIC)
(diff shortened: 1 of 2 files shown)


ppisar:
I also corrected a python test.

m-blaha (approved):
Thank you!

PR #322, opened 2024-07-19
Fix memory leaks when using zchunk

kontura:


librepo/downloadtarget.c
```diff
@@ -134,6 +134,26 @@ lr_downloadtarget_free(LrDownloadTarget *target)
     if (!target)
         return;
 
+    #ifdef WITH_ZCHUNK
+    if (target->zck_dl) {
+        zckCtx *old_zck = zck_dl_get_zck(target->zck_dl);
```
jan-kolarik, line 139: Is't just a minor thing, but is there a reason why refer to the variables in this function as `old_*`? I don't see it being replaced by some other or new values in this context...

jan-kolarik: I just want to make sure I am not missing something ...

kontura: No reason, I just copied it.

Should I change it? It's no problem.

jan-kolarik: OK, if it's a matter of few minutes, then yes please. Better to avoid any potential confusion when reading the code.

kontura: Updated.

There appear to be some copr connectivity problems in the CI.

jan-kolarik (approved):

PR #325, opened 2024-10-02
Fix input termination for `pgpParsePkts`

jrohel:
The `pgpParsePkts` function needs the OpenPGP ASCII armored input to be null terminated. The librepo contains code that checks if the input is null-terminated. If it is not, the code creates a local null-terminated copy of the input.

There was a bug in the code, so it may look for a terminating null several bytes behind the input buffer. And when a null was found behind the input buffer, the termination was not done. This caused the `pgpParsePkts` function to process several extra characters after the input buffer. These characters are generally random and sometimes cause the `pgpParsePkts` function to return an error.

I hope this fixes https://github.com/rpm-software-management/dnf/issues/2107

### DIFF
--- a/librepo/gpg_rpm.c
+++ b/librepo/gpg_rpm.c
@@ -350,7 +350,7 @@ lr_gpg_import_key_from_memory(const char *key, size_t key_len, const char *home_
 
     // `pgpParsePkts` needs null-terminated input, if null byte not found, make a local null-terminated copy
     g_autofree gchar * key_with_null_byte = NULL;
-    if (memchr(block_begin, '\0', key_len) == NULL) {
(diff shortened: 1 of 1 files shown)


m-blaha (approved):
Great catch! This is one of those "easy to miss" errors.

praiskup:
Any plan to wrap a release?

evan-goode:
Updates are pending :)

https://bodhi.fedoraproject.org/updates/FEDORA-2024-5cce41cf3e
https://bodhi.fedoraproject.org/updates/FEDORA-2024-c912f70280
https://bodhi.fedoraproject.org/updates/FEDORA-2024-32c01ef78b
https://bodhi.fedoraproject.org/updates/FEDORA-2024-586ce2a0a9

PR #331, opened 2024-11-07
Set CMAKE_SKIP_RPATH to TRUE

bluca:
When skipping RPATH make doc raise:

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/lib/python3/dist-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "/home/user/librepo/doc/python/conf.py", line 15, in <module>
    import librepo
  File "/home/user/librepo/build/librepo/python/python3/librepo/__init__.py", line 1081, in <module>
    import librepo._librepo
ImportError: librepo.so.0: cannot open shared object file: No such file or directory
```

A fix is to set properly LD_LIBRARY_PATH in sphinx env call


bluca:
> Why do you want to do this?

Because otherwise building the docs fails, as mentioned in the commit message

PR #332, opened 2024-11-07
Prefer pkgconf to find gpgme

bluca:
gpgme-config has been deprecated and is not available in some distributions such as Debian, so prefer pkg-config and only use it as a fallback.


Conan-Kudo (approved):
This should be fine now, since we don't support EL8 anymore.

Issue #251, opened 2021-07-08
RFE: Support for requesting metalink via header on baseurl


Conan-Kudo:
openSUSE's mirror infrastructure is set up so that the download redirector will return metalink information upon request for each file when `"Accept: */*, application/metalink+xml, application/metalink4+xml"` is passed in as a header.

This is used by libzypp to be able to selectively metadata from the primary server and fetch packages from mirrors with failover. We should have support for this too.

DemiMarie:
If I turn off metalinks (by using baseurl), I expect none of the metalink code to run.  The reason I would turn off metalinks is the reduced attack surface.

DemiMarie:
> Yes, but the idea with this feature is that the `repomd.xml` and `repomd.xml.asc` files can be fetched from the master server, and all the RPMs can be fetched via metalink. That can be very useful from a security perspective without punishing a server's allotted bandwidth.

That is an improvement over using metalinks for everything, yes, but I would prefer if the metalinks were statically-generated and signed.  This might require a Debian-style CDN, though.  I am specifically worried about libxml2, which is a rather large C library.

PR #336, opened 2025-03-18
Fix `lr_download_metadata(...)` API 

kontura:
Required for: https://github.com/rpm-software-management/dnf5/issues/307

Required by https://github.com/rpm-software-management/dnf5/pull/2167


pkratoch:
I am not a good reviewer for this. I did my best to understand it and can say that it looks sane, but I cannot be very confident about this, because I don't know librepo that well. We could either consider asking someone more familiar with this to take a look or at least wait for the review of https://github.com/rpm-software-management/dnf5/pull/2167 which uses this one and might discover potential issues.

Conan-Kudo:
The code looks reasonable to me, but I'm not sure what's up with the failing CI tests as GitHub doesn't seem to want to give me information about it.

Conan-Kudo:
Question, how would this fit alongside eventually implementing #251 to have us fetch the `repomd.xml` and `repomd.xml.asc` file from the main server while getting everything else from mirrors?

kontura:
> The code looks reasonable to me, but I'm not sure what's up with the failing CI tests as GitHub doesn't seem to want to give me information about it.

I believe they are due to https://github.com/rpm-software-management/dnf5/issues/2176.
I will rerun the tests once the unrelated problems are resolved.

kontura:
> Question, how would this fit alongside eventually implementing #251 to have us fetch the `repomd.xml` and `repomd.xml.asc` file from the main server while getting everything else from mirrors?

I think it doesn't change much in that regard. It still uses the same mechanisms and logic as before it just changes the order.
So with or without this PR I believe the changes needed for https://github.com/rpm-software-management/librepo/issues/251 will stay the same.

kontura:
Upon further consideration I have decided to drop the problematic commit https://github.com/rpm-software-management/librepo/commit/97089b0ddb66f82d5613fb903857d3ce858ea86d. 
In essence it seems to me that a repo with no (or invalid) mirrors configured (`baseurl`, `mirrorlist` or `metalink`)  should be always reported as an error regardless of `skip_if_unavailable` option.

Actually if there is no mirror `libdnf5` won't even pass such a repo to librepo.
The error from the mentioned commit can happen only if an invalid url is provided such as: 
`dnf5 --repofrompath=test,asdasd --repo test rq`. Librepo will test `asdasd` using `realpath.1` and if it doesn't exist it will report the `Empty mirrorlist and no basepath specified!` error.
I don't think we should treat such a repo as unavailable (and allow skipping it), in my opinion it is more likely an error in configuration (and stopping all downloads is appropriate).

(Note that the CI should now correctly test this code.)

Conan-Kudo (approved):

Issue #303, opened 2024-03-08
"DNF4 Integration Tests" CI test does not use librepo built from a pull request


ppisar:
In PR #302 we noticed that "DNF4 Integration Tests" tests a wrong code on a pull request. Instead of installing librepo built from the pull request, it installs librepo from copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly repository. That repository contains a day old snapshot.

That means the "DNF4 Integration Tests" is never run for a pull request code and completely misses the point of CI.

The same applies to "DNF5 Integration Tests" task.

kontura:
This is resolved by https://github.com/rpm-software-management/librepo/pull/337.
Both dnf4 and dnf5 tests now use the correct code from the given PR.

PR #337, opened 2025-05-06
Fix CI by passing the built packages urls

kontura:
The job names were incomplete which led to empty package-urls so the code from the given PR was never tested.

For: https://github.com/rpm-software-management/librepo/issues/303

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
       - name: Run Integration Tests
         uses: ./.github/actions/integration-tests
         with:
-          package-urls: ${{needs.copr-build.outputs.package-urls}}
+          package-urls: ${{needs.copr-build-dnf4.outputs.package-urls}}
 
   integration-tests-dnf5:
     name: DNF5 Integration Tests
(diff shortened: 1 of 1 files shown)


ppisar:
The change is logical, but does it help? Reading the [DNF4 Integration Tests](https://github.com/rpm-software-management/librepo/actions/runs/14852601561/job/41699249558?pr=337#logs) output I can see that it installs:

~~~~
librepo                      x86_64  1.19.0-20250506011204.4.g8a68902.fc40            copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf-nightly
~~~~

That means it does not use librepo built from this commit.

kontura:
> That means it does not use librepo built from this commit.

Yes, this is because it takes the workflow definition from the base branch of the PR (`master`). I believe this is a security feature. Additional details can be found here: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target (we use the `pull_request_target` event).

To verify it works I have setup up my forked master branch which already contains the change and I made a testing PR: https://github.com/kontura/librepo/pull/4. We can see either in the actions output or directly in the tests results that it in fact used the new code to the testing.

ppisar (approved):
I see. Though I still find the security schema weird: I would understand halting the CI and waiting on an approval of the maintainer. Running CI on a different code is not helpful, I think.

PR #338, opened 2025-05-28
Parallel downloading of metalinks/mirrorlists and fetchmirrors and update

kontura:
Downloading in parallel is done in several stages:
1. All metalinks and mirrorlists are downloaded in parallel
2. All repomds are downloaded in parallel
3. All remaining repo metadata (primary, filelists,...) are downloaded in parallel

Ideally we would combine 1. and 2. into a single parallel download (this is what libdnf needs to determine if repo is up to date), but it would be quite a complex change to how librepo works.

### DIFF
--- a/librepo/metadata_downloader.c
+++ b/librepo/metadata_downloader.c
@@ -298,11 +298,10 @@ create_repomd_xml_download_targets(GSList *targets,
 }
 
 // If it returns FALSE then err is set
-gboolean
+void
 process_repomd_xml(GSList *targets,
                    GSList *fd_list,
-                   GSList *paths,
-                   GError **err)
+                   GSList *paths)
 {
     GError *error = NULL;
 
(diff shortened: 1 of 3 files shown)


librepo/metadata_downloader.c
```diff
@@ -465,6 +462,96 @@ restore_handle_callbacks(GSList *targets, GSList *handle_callbacks_backups)
...
+
+        if (handle->metalinkurl) {
+            url = lr_prepend_url_protocol(handle->metalinkurl);
+        } else if (handle->mirrorlisturl) {
+            url = lr_prepend_url_protocol(handle->mirrorlisturl);
+        } else {
+            // Creating targets only for metalinks or mirrorlists
+            assert(false);
+        }
+        handle->onetimeflag_apply = TRUE;
```
evan-goode, line 501: There might be an issue here... CI test `dnf/countme.feature:124  Countme flag is not sent repeatedly on retries` is failing (ran locally with https://github.com/rpm-software-management/dnf5/pull/2210).

kontura: Oh, I have missed that test.
I found out the problem is not here but in how many retries are made if the mirror is failing. I have updated the PR with fix and a comment.
I have also noticed the original (non-parallel) code path is downloading both mirrorlist and metalink if they are both available so I made additional changes to do that also in the parallel downloads.

evan-goode (approved):
Thanks!

PR #343, opened 2025-06-16
test: Use 1 instead of true

ppisar:
Building in RHEL 10.0 failed:

    /home/test/librepo/examples/c/download_repos_parallel.c: In function ‘create_handle’:
    /home/test/librepo/examples/c/download_repos_parallel.c:52:48: error: ‘true’ undeclared (first use in this function)
       52 |   lr_handle_setopt(h, NULL, LRO_INTERRUPTIBLE, true);
          |                                                ^~~~

true is not implicitly defined until ISO C23.

This patch fixes it by using a numerical value, similar to commit 0dae97aafc4d0112cc192dd6c9b74687342431f5 ("Use 0 instead of `false` for assert").

### DIFF
--- a/examples/c/download_repos_parallel.c
+++ b/examples/c/download_repos_parallel.c
@@ -49,13 +49,13 @@ LrHandle *create_handle(const char *repo) {
 
   // --- Optional arguments --------------------------------------------
   // Make download interruptible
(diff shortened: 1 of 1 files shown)


kontura:
Thank you!

Issue #288, opened 2023-10-16
librepo fails to build in dnf-nightly COPR for RHEL 8


ppisar:
I notices that dnf-nightly COPR repository attempts to build librepo for CentOS Stream 8 and EPEL 8 <https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/package/librepo/>. Building a master branch for RHEL 8 is pointless because master librepo requires glib2 ≥ 2.66 while RHEL 8 has older 2.56.4 version.

I recommend reconfiguring librepo package in the COPR repository not to build it in RHEL 8 derivates.

PR #344, opened 2025-06-17
Add packit copr-builds and tests

kontura:
Also adds a epel copr-build targets to test building there as well.


kontura:
I believe the dnf5 testing failures are caused by: https://github.com/rpm-software-management/ci-dnf-stack/pull/1687

While I added the epel builds I am not sure we want them here.

.packit.yaml
```diff
@@ -0,0 +1,51 @@
...
+# This is needed because other packages can depend on the new version.
+actions:
+  get-current-version:
+    - rpmspec --srpm --query --queryformat "%{version}" librepo.spec
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      - epel-all
```
ppisar, line 16: Exclude EPEL-8. librepo from main branch requires newer glib2 (issue #288).

PR #345, opened 2025-06-24
rhel-9.7: Propagate return value from `prepare_repo_download_targets`

kontura:
Backport: https://github.com/rpm-software-management/librepo/pull/341


ppisar (approved):
Good.

PR #352, opened 2025-09-24
Propagate return value from `prepare_repo_download_targets`

kontura:
It also ensures that when return value is false there is some error set.

Backport: https://github.com/rpm-software-management/librepo/pull/341


ppisar (approved):
Good.

Issue #207, opened 2020-11-02
Properly handle expired GPG signatures


DemiMarie:
This was the subject of a CVE in APT.

PR #354, opened 2025-10-10
Introduces a proper handle expired GPG signatures - GPGME backend

fhbash:
This change only affects the gpgme implementation of signature verification.
The rpm implementation (gpg_rpm.c) is not modified as it already rejects expired keys during import.

Changes:
- Implements proper detection and handling of expired GPG keys and signatures   in the gpgme backend (gpg_gpgme.c)
- Adds support for key rotation: if data is signed with multiple keys and  at least one valid, non-expired signature exists, verification succeeds
- Only fails verification if ALL signatures are either invalid or expired
- Provides clear error messages distinguishing between expired keys and  expired signatures
- Adds test case to verify expired signature rejection

type: enhancement
resolves partially: https://github.com/rpm-software-management/librepo/issues/207

**Which this PR not resolve**:  checking using RPM backend
**Why**: current rpm api doesn't provide an api to check multiple sign/key, RPM team already have and ticket for that and currently not merged, for more details, see: https://issues.redhat.com/browse/RHEL-112394

I created new issue to deal with rpm backend as soons the change on rpm is deployed. Here it is the ticket: https://github.com/rpm-software-management/librepo/issues/356

Signed-off-by: Fellipe Henrique <me@fhbash.com>


ppisar:
The packit failure was probably worked around in #360. I will rebase your pull request to make that workaround effective here.

PR #355, opened 2025-10-20
Add build requirements installation instructions

fhbash:
This is an small change, adding instructions for installing build requirements directly from spec file


ppisar (approved):
Good.

PR #357, opened 2025-10-24
Add support for excluding domains and locations in metalink parsing

bc-lee:
This commit introduces two new options, LRO_METALINK_EXCLUDE_DOMAIN and LRO_METALINK_EXCLUDE_LOCATION, to the LrHandle structure. These options allow users to specify lists of domains and locations to exclude during metalink processing.

Related: https://github.com/rpm-software-management/dnf5/issues/1483


bc-lee:
Thanks for the review. I think I need to redesign the API to reduce API/ABI incompatibilities. It will take a few days, so I’ll mark this PR as a draft for now.

PR #361, opened 2025-11-12
Do not fail in lr_prepare_repodata_dir() if repodata/ dir exists

dcantrell:
If mkdir() fails and errno is EEXIST, that's ok.  It means we already have the repodata/ directory.  Just keep going.  But any other errors get reported and the function fails.

Fixing this fixes the problem I've been seeing on my system ever since moving to dnf5, which is:

    $ sudo dnf check-update
    # ERROR: something about directory already existing
    # argh..., so then do
    $ sudo dnf clean all
    $ sudo dnf check-update

Now I can just do 'sudo dnf check-update' and it works.  I am seeing this on Fedora 42 right now, but I also saw it on previous releases.


fhbash (approved):
LGTM

Conan-Kudo (approved):

Issue #293, opened 2024-01-01
DNF5 always removes partially downloaded files when interrupted, so you cannot resume downloading them


hedayat:
I tried both running `dnf5 install ...` and `dnf5 download ...` in Fedora 39, and interrupting them in the middle of downloading a large package. Upon interruption, the partially downloaded file is removed. Hence, you cannot resume downloading the package. I tried also adding `keepcache` option, but it has no effect.

Effectively, DNF5 doesn't have support for resume functionality right now.

jan-kolarik:
Hi, I've been testing the described issue on my local machine, and it appears to be related to the librepo component handling package downloads. Therefore, I've transferred the problem there.

The issue lies in how `librepo` processes interruptions. When a `SIGTERM` is sent to the DNF process, the partially downloaded file is retained and resumed upon the next download attempt. However, when a `SIGINT` (<kbd>CTRL</kbd>+<kbd>C</kbd>) is sent, the file is discarded, resulting in the following error:

```
>>> Not finished - interrupted by error: Interrupted by signalFailed to download packages
 Librepo error: Interrupted by a SIGINT signal
```

Upon comparing this behavior to DNF4, I observed that it behaves similarly, just with different output.

jan-kolarik:
I wasn't investigating deeply why the file is removed when `SIGINT` is used, but it seems that the issue may be related to this section: https://github.com/rpm-software-management/librepo/blob/master/librepo/downloader.c#L2817.

hedayat:
> Upon comparing this behavior to DNF4, I observed that it behaves similarly, just with different output.

TBH, I hope DNF5 doesn't try to replicate bad behaviors from DNF4 :P Although, I agree it depends on how you interpret Ctrl+C: Cancel vs Interrupt.

But, if the current behavior is retains, so we should either terminate DNF5 by other means or hope it'll terminate by itself upon timeout/...? 

And, does it mean that if  DNF has downloaded a number of packages and you press Ctrl+C, it'll remove all of them? Not sure that's what a user expects :thinking:

jan-kolarik:
> TBH, I hope DNF5 doesn't try to replicate bad behaviors from DNF4 :P Although, I agree it depends on how you interpret Ctrl+C: Cancel vs Interrupt.

> But, if the current behavior is retains, so we should either terminate DNF5 by other means or hope it'll terminate by itself upon timeout/...?

[...]

PR #369, opened 2026-03-20
Fix download resume within a single session

m-blaha:
Download resume was effectively broken - when a transfer failed and was retried on another mirror, the partial data was always discarded and the download restarted from scratch. Several issues contributed to this:

1. Header callback breaks resumed transfers: For HTTP 206 Partial Content responses, Content-Length contains the remaining bytes, not the full file size.

2. Retry always discards partial data: On transfer failure, truncate_transfer_file() unconditionally truncated back to original_offset (0 for new files), and original_offset was never updated. Now, for connection errors (timeout, recv error, etc.) where partial data is likely valid, the file is kept and original_offset is updated to the current file size so the next attempt resumes from where it left off. Server errors (bad HTTP status), checksum failures, and header callback interrupts still truncate as before.

3. File position not set on resume: When resuming a download, the file pointer was left at position 0 after fdopen. Data received from the server (starting at the resume offset) was written at position 0, corrupting the file.

4. resume flag not re-enabled on retry: The xattr check in prepare_next_transfer permanently disabled the resume flag for new files (no xattr on first open). On retry after a connection error, resume was not re-enabled, preventing the use of CURLOPT_RESUME_FROM_LARGE.

5. resume_count incremented on non-resume attempts: The resume attempt counter was incremented on every prepare_next_transfer call, even for fresh downloads (offset 0). This meant resume was disabled before any actual resume could happen.

Also a test is added for resuming the download after a mirror switch.

Related: https://github.com/rpm-software-management/librepo/issues/293

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -193,6 +193,9 @@ typedef struct {
 
     gboolean range_fail; /*!<
         Whether range request failed. */
+
+    CURLcode curl_code; /*!<
+        Result code from the last curl transfer */
 } LrTarget;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


librepo/downloader.c
```diff
@@ -2449,17 +2472,39 @@ check_transfer_statuses(LrDownload *dd, GError **err)
...
+                        // Connection error (timeout, recv error, etc.) with
+                        // potentially valid partial data. Keep the data and
+                        // set offset to current file size so the next attempt
+                        // resumes from where we left off.
+                        if (target->target->fn) {
+                            struct stat st;
+                            if (stat(target->target->fn, &st) == 0) {
+                                target->original_offset = st.st_size;
+                            }
+                        }
```
kontura: It is quite likely I am missing something but similar logic is already present at https://github.com/rpm-software-management/librepo/blob/master/librepo/downloader.c#L1586-L1596 what is the reason to have it here again?
It is present in `prepare_next_transfer` which should be called at the end here (in `check_transfer_statuses`).

It is guarded by `if (target->original_offset == -1) {` but that condition seems very suspicious to me. As far as I can tell `original_offset` is only initialized to -1 and its immediately set to 0 in https://github.com/rpm-software-management/librepo/blob/master/librepo/downloader.c#L2750 before any download starts.

Could it be further unified?

kontura (commented):
These are some nice finds and fixes!

librepo/downloader.c
m-blaha: Oh, right. Great observation! We can just reset `target->original_offset = -1` here, and let the prepare_next_transfer recalculate the actual position for the resumed download. That would nicely simplify the patch.

kontura (approved):
Thank you!

PR #370, opened 2026-03-27
Configure packit to use caret notation for postrelease snapshots

kontura:
This ensures the build of that specific version is considered newer than any other regular release of it.
For https://github.com/rpm-software-management/ci-dnf-stack/issues/1843

Example build: https://copr.fedorainfracloud.org/coprs/amatej/dnf-nightly/build/10266049/


m-blaha (approved):
Let's try it!

PR #371, opened 2026-03-30
Enable HTTP/2, connection reuse, and improve download parallelism

aruiz:
## Summary

- Default all transfers to HTTP/2 over TLS (`CURL_HTTP_VERSION_2TLS`) with transparent fallback to HTTP/1.1. Expose `LRO_HTTPVERSION` option as an escape hatch.
- Add a `CURLSH` share handle per `LrHandle` to pool DNS cache and TCP+TLS connections across sequential `lr_download()` calls, avoiding redundant handshakes.
- Raise default max parallel downloads from 3 to 5 and fix `lr_download()` to use the maximum parallelism across all target handles rather than just the first one.

## Results

Benchmarked with `dnf5 --setopt=build_cache=0 --refresh makecache` against 9 enabled repos (Fedora 41, RPM Fusion, Chrome, VS Code, etc.):

| | Median | Mean |
|---|---|---|
| Before (system librepo) | 3.78s | 3.85s |
| After (patched librepo) | 2.43s | 2.49s |
| **Speedup** | **36%** | **35%** |

No changes required in dnf5 — all improvements take effect transparently.

## Disclosure

This changeset was developed with assistance from Claude (Anthropic). Each commit is tagged `Assisted-by: Claude`.


Conan-Kudo:
Everything is failing with this PR, could you please take a look?

Conan-Kudo:
This needs rebasing, and also it seems like your Git author is unknown to GitHub?

PR #375, opened 2026-06-12
Close zchunk seed fd in find_local_zck_header

m-blaha:
`find_local_zck_header()` opens the old cached .zck file as a seed for delta downloads but never closes the fd when a match is found (the break skips the close call). The fd stays open until `lr_free(target)`, which happens after the end callback.

dnf5 registers an end callback that removes the old repodata/ directory and moves freshly downloaded metadata into place. On NFS, unlinking a file with an open fd does not remove the directory entry but creates a .nfs* silly-rename file instead. The final rmdir then fails with ENOTEMPTY, causing "Failed to download metadata" errors.

The seed file content is already fully copied to the target fd via `lr_copy_content()` before `lr_zck_init_read()` is called. Pass the target fd instead of the seed fd to `lr_zck_init_read()` so that the seed fd can be closed immediately.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2459175

### DIFF
--- a/librepo/downloader.c
+++ b/librepo/downloader.c
@@ -1103,7 +1103,8 @@ find_local_zck_header(LrTarget *target, GError **err)
                    ftruncate(fd, lseek(chk_fd, 0, SEEK_END)) >= 0 &&
                    lseek(fd, 0, SEEK_SET) == 0 &&
                    (zck = lr_zck_init_read(target->target, (char *)file->data,
-                                           chk_fd, &tmp_err))) {
+                                           fd, &tmp_err))) {
+                    close(chk_fd);
                     found = TRUE;
                     break;
                 } else {


librepo/downloader.c
```diff
@@ -1103,7 +1103,8 @@ find_local_zck_header(LrTarget *target, GError **err)
                    ftruncate(fd, lseek(chk_fd, 0, SEEK_END)) >= 0 &&
                    lseek(fd, 0, SEEK_SET) == 0 &&
                    (zck = lr_zck_init_read(target->target, (char *)file->data,
```
ppisar, line 1105: The fix makes sense. But what about the file name passed as a second argument to lr_zck_init_read()? Is it correct? file->data is a name of chk_fd descriptor. I can see that chk_fd and fd have the same content, but still lr_zck_init_read() operates on fd. Should the file name argument be replaced with target->target->fn?

I can see that struct LrDownloadTarget definition documents that either fn, or fd is set, not both. But `target->f = open_target_file(target, err);` in prepare_next_transfer() which opens fd does not unset fn.

m-blaha: I believe the current (`file->data`) filename is correct. The `lr_zck_init_read()` function doesn't store the filename anywhere and uses it exclusively for debug messages. Here, I think, the original *.zck filename (e.g. `/var/cache/libdnf5/fedora/repodata/primary.xml.zck`) makes more sense than the target (which is located in a temp directory and later moved).

ppisar (approved):
The dnf4 CI failure ("non-UTF-8 character in an option when using corresponding locale") is unrelated <https://github.com/rpm-software-management/ci-dnf-stack/issues/1893>.

PR #377, opened 2026-06-19
Fix metadata file descriptor leaks

kontura:
The opened `fd`'s were never closed.

We don't have to preopen `fd` for `LrMetadataTarget` rather pass it by `fn`. It is later correctly opened with the same flags (taking zchunk into account) and closed.

Discovered in dnf5 by running: `valgrind --track-fds=yes ./build/test/libdnf5/run_tests`
(There is still one fd leak but its a test implementation problem that is unrelated.)

### DIFF
--- a/librepo/yum.c
+++ b/librepo/yum.c
@@ -669,22 +669,9 @@ gboolean
 prepare_repo_download_std_target(LrHandle *handle,
                                  LrYumRepoMdRecord *record,
                                  char **path,
-                                 int *fd,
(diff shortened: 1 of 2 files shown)


ppisar (approved):
Good. The change looks reasonable and dnf5 tests now do not leak this file descriptor.

ppisar:
The failures in F45 are unrelated (merging OpenSSL 4 which added openssl3 compat package).

PR #378, opened 2026-06-23
Unify error messages for GPG errors across backends

### LINKED ISSUE #376: librepo rejects valid repomd.xml GPG signatures as "Bad GPG signature" with gpgme 2.1.0
## Summary

`librepo` fails repository **metadata** signature verification (`repo_gpgcheck`) with

```
repomd.xml GPG signature verification error: Bad GPG signature
```

for a `repomd.xml` whose detached signature is in fact **valid**. The `gpg(1)` CLI and `gpgme` (the library librepo itself uses) both verify the *same bytes* against the *same key* successfully — only librepo rejects it.

This breaks `dnf5` (which uses librepo) when consuming any repository that signs its metadata (e.g. CentOS Stream BaseOS/AppStream/CRB). It surfaced when building CentOS images with [mkosi](https://github.com/systemd/mkosi) using an openSUSE or Arch tools tree.

I originally reported this against openSUSE in https://bugzilla.opensuse.org/show_bug.cgi?id=1268271 but now I saw the [same issue on Arch](https://github.com/systemd/mkosi/actions/runs/27537658400/job/81400133285?pr=4361).

Arch has gpg 2.4.9 (same as Fedora 44, where it's working), gpgme 2.1.0 (same as openSUSE, where it's failing -- Fedora 44 has 2.0.1 which works), and librepo 1.20.0 (same across all three distributions). So this was triggered by updating gpgme from 2.0.1 to 2.1.0.

## What each layer reports (same bytes, same key)

- **`gpg(1)` CLI:** `Good signature`
- **`gpgme` (python3 bindings, fresh homedir + gpgme key import — as librepo does):**
  `status = GPG_ERR_NO_ERROR` (valid), `summary = 0x0`
- **`librepo` / `dnf5`:** `Bad GPG signature`, operation fails (exit 1)

gpgme reports the signature as cryptographically valid (`status == 0`); `summary == 0` only means the key is untrusted, which is normal/expected for repository verification. So the discrepancy appears to be in **librepo's interpretation of the gpgme verification result**,
e.g. treating "valid signature from an untrusted key" (`summary` without `GPGME_SIGSUM_VALID`) as a bad signature, or a mismatch with a gpgme 2.x API/behavior change.

The key is the CentOS Official signing key (RSA-4096, key id `05B555B38483C65D`); the signature is a v4 RSA/SHA-256 signature.

## Reproducers

I made two reproducer scripts: One for Arch: [repro-arch-repomd-gpg.sh](https://github.com/user-attachments/files/28953389/repro-arch-repomd-gpg.sh) and one for openSUSE: 

[repro-suse-repomd-gpg.sh](https://github.com/user-attachments/files/28953393/repro-suse-repomd-gpg.sh)

Each fetches one internally-consistent `repomd.xml` + `repomd.xml.asc` from the canonical CentOS Stream mirror and verifies the same bytes against the same key three ways. They can be run on a pristine arch/tumbleweed OS, but the easiest is probably to run them in a container:

```
podman run --rm -v $PWD/repro-suse-repomd-gpg.sh:/repro.sh:ro docker.io/opensuse/tumbleweed bash /repro.sh
```

```
podman run --rm -v $PWD/repro-arch-repomd-gpg.sh:/repro.sh:ro docker.io/archlinux bash /repro.sh
```

They both show that gpg CLI and gpgme (via Python) work fine, but dnf5/librepo fail. Slightly abbreviated output:

```
## (A) gpg(1) CLI  -- verifies the detached signature
gpg: key 05B555B38483C65D: public key "CentOS (CentOS Official Signing Key) <security@centos.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: Signature made Mon Jun  1 14:51:56 2026 UTC
gpg:                using RSA key 05B555B38483C65D
gpg: Good signature from "CentOS (CentOS Official Signing Key) <security@centos.org>" [unknown]

## (B) gpgme (python-gpg) -- fresh homedir + gpgme key import, like librepo does
   summary=0x0 status=OK  (status OK == cryptographically valid)

## (C) dnf5 / librepo -- same bytes, same key
>>> repomd.xml GPG signature verification error: Bad GPG signature              
Failed to download metadata (baseurl: "https://mirror.stream.centos.org/10-stream/BaseOS/x86_64/os/") for repository "baseos": repomd.xml GPG signature verification error: Bad GPG signature
```

m-blaha:
When `repo_gpgcheck` is enabled, dnf5 first attempts to verify the repository metadata signature and, if the error message indicates the signing key is not in the keyring (`Signing key not found`), it downloads and imports the key from gpgkey URLs and retries.

The gpgme backend did not distinguish between a missing key and a genuinely bad signature, and reported the generic "Bad GPG signature" for both cases. This prevented dnf5's key import retry from triggering, causing repo_gpgcheck=1 to always fail on the first run on distros that build librepo with `USE_GPGME=ON` (Arch Linux, openSUSE).

Also a test is added to verify, that a missing key is correctly reported for backend used to compile librepo.

Resolves: https://github.com/rpm-software-management/librepo/issues/376

### DIFF
--- a/librepo/gpg_gpgme.c
+++ b/librepo/gpg_gpgme.c
@@ -286,8 +286,13 @@ lr_gpg_check_signature_fd(int signature_fd,
     gboolean found_valid_signature = FALSE;
     gboolean found_expired_key = FALSE;
     gboolean found_expired_sig = FALSE;
+    gboolean found_missing_key = FALSE;
     for (; sig; sig = sig->next) {
-        // Check if this signature is from an expired key or is itself expired
+        if (sig->summary & GPGME_SIGSUM_KEY_MISSING ||
+            gpg_err_code(sig->status) == GPG_ERR_NO_PUBKEY) {
+            found_missing_key = TRUE;
+            continue;
+        }
         if (sig->summary & GPGME_SIGSUM_KEY_EXPIRED) {
             g_debug("%s: Skipping signature with expired key", __func__);
             found_expired_key = TRUE;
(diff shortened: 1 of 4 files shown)


evan-goode (approved):
Great, very clearly written, thanks!

PR #387, opened 2026-09-10
downloader: don't rescan every target when nothing can be started

stewartsmith:
select_next_target() walks the list of targets looking for one that is waiting and whose mirrors can take another connection. Two things made that walk run far more often, and cost far more, than it needs to.

check_transfer_statuses() runs after every curl_multi_perform() and finished by calling prepare_next_transfers() unconditionally, even when no transfer had finished. Nothing can have changed in that case: targets only change state, and mirror connections are only released, when a transfer completes. So the walk could only reach the same conclusion as the one before it.

Worse, when maxparalleldownloads is larger than maxdownloadspermirror there is permanently a free transfer slot that no mirror is able to accept. Every one of those walks therefore examined every single waiting target, calling select_suitable_mirror() and logging for each of them, before concluding that nothing can start. The cost of one walk grows with the number of waiting targets, which makes downloading a whole repository quadratic.

Ask for the next transfers only once a transfer has actually finished, and skip targets whose mirrors have all reached their connection limit without consulting those mirrors one target at a time.

Downloading 512 byte files over loopback, so that what is measured is librepo's own bookkeeping rather than the network, before -> after:
```
  targets   parallel/per-mirror            wall                 user
    5,000          20 / 3          8.29s ->   1.96s     8.75s ->   1.64s
   20,000          20 / 3        117.54s ->  10.88s   117.26s ->  10.53s
    5,000           3 / 3          1.89s ->   1.89s     1.64s ->   1.46s
    5,000          20 / 20         2.61s ->   2.62s     1.84s ->   1.70s
```
Four times the targets costs 14.2 times the wall time before the change and 5.6 times after it. Configurations where the two limits are equal, which includes the defaults, are unaffected. Every target downloaded successfully in every run, before and after.

Note that libdnf5 exposes the two limits as independent options that both default to 3, so raising only max_parallel_downloads - the obvious way to try to speed downloads up - is what lands in the slow case.


m-blaha (approved):
Nice, thank you!

PR #388, opened 2026-09-11
downloader: hold transfer state per connection, not per target 

stewartsmith:
**NOTE:** This is on top of https://github.com/rpm-software-management/librepo/pull/387 so it's probably best to just look at the top level commit here as github will display the diff including https://github.com/rpm-software-management/librepo/pull/387 here.

---

lr_download() allocates one LrTarget for every target up front and keeps it
until the whole batch finishes. Most of that structure was not target state
at all but the resources and scratch of a single transfer attempt - the curl
handle, the FILE, the 256-byte CURL_ERROR_SIZE buffer, the request headers,
the protocol and the header/write callback scratch. At most
LRO_MAXPARALLELDOWNLOADS transfers run at a time, so a caller handing
librepo a whole repository paid for thousands of copies of state that only
a handful of connections could ever use.

Move that state into a new LrTransfer, allocated in prepare_next_transfer()
when a transfer starts and freed in transfer_end() when it finishes.
LrTarget goes from 400 to 96 bytes, with one 312-byte LrTransfer per running
transfer rather than one per target: for a 50k-package batch at the default
3 parallel downloads, about 20.8 MB of LrTargets becomes about 5.6 MB plus
one kilobyte.

Membership is decided by whether anything reads a field after the transfer
has been torn down. The verdict of an attempt - curl_code, headercb_state,
original_offset, cb_return_code, zck_state, range_fail - is consumed by the
retry logic in check_transfer_statuses() after teardown, so those stay on
LrTarget. The error buffer is safe to release because
check_finished_transfer_status() has already turned it into a GError, the
same discipline the effective URL already relies on.

The four places that ended a transfer each open-coded the teardown; they now
share transfer_end(). That incidentally fixes a leak of curl_rqheaders when
a batch was abandoned by an error, which only one of the four handled.

Tested with -DWITH_ZCHUNK=ON and -DWITH_ZCHUNK=OFF: 51/51 checks pass in
both, and valgrind reports no invalid accesses and no leaks anywhere in
downloader.c (the one leak the suite still reports is a pre-existing one in
test_gpg.c). librepo has no test that drives a zchunk download, so the
zchunk paths here are compile-checked but not exercised at runtime.


m-blaha (approved):
Thank you, this is a nice memory optimization!

PR #390, opened 2026-09-13
handle: Validate the new value of LRO_MAXMIRRORTRIES

prownd:
lr_handle_setopt() compared the currently stored value (handle->maxmirrortries) against LRO_MAXMIRRORTRIES_MIN instead of the value that is being set. As the stored value is always valid, the check never triggered and any value, including negative ones, was accepted.

Check the value passed by the caller and fix the error message, the minimal value itself is allowed.

Add a unit test.


m-blaha (approved):
Thank you!

PR #391, opened 2026-09-13
xmlparser: Check the parser context before using it

prownd:
- lr_xml_parser_generic() dereferenced the result of xmlCreatePushParserCtxt() before the assert() that checks it for NULL. Report an error instead of crashing when the context can't be created.
- lr_char_handler() used plain realloc() on a buffer allocated by g_malloc() and didn't check the result. Use g_realloc(), which aborts on OOM consistently with the rest of the code.
- lr_xml_parser_strtoll() checked errno without resetting it first, so a stale ERANGE could produce a bogus overflow warning.


m-blaha (approved):
Thank you!

m-blaha:
I was just wondering whether it would make sense to split the fix into three commits, since it addresses three different issues. However, given how small the fixes are, I’m fine with keeping them in a single commit.

PR #392, opened 2026-09-13
tests: Run the repo_zck test suite

prownd:
The suite was added together with the zchunk support, but it was never registered in test_main.c, so it has never run. It also expects that repo_yum_03/repodata/repomd.xml has 12 records, while it has 6.

Register the suite and fix the expected number of records.


m-blaha (approved):
Thank you!

PR #393, opened 2026-09-13
downloader: Compare HTTP header names case-insensitively

prownd:
HTTP header field names are case-insensitive (RFC 9110, section 5.1) and HTTP/2 and HTTP/3 servers always send them in lower case, so lr_headercb() never found "Content-Length: " there and silently skipped the check of the size reported by the server against the expected size (a file of a wrong size was then accepted if there was no checksum to verify).

Compare the header name case-insensitively and don't require exactly one space after the colon.


m-blaha:
Thanks for the PR! The code looks good. Would you mind adding a regression test for this change?

prownd:
You're right.
A new one should be added.

prownd:
> Thanks for the PR! The code looks good. Would you mind adding a regression test for this change?

Added.

tests/test_downloader.c
```diff
@@ -373,6 +378,83 @@ START_TEST(test_downloader_checksum)
...
+
+    pid_t server = fork();
+    ck_assert_int_ge(server, 0);
+    if (server == 0)
+        http_server(sock, response);
+    close(sock);
+
+    // Don't let a proxy from the environment handle the transfer
+    unsetenv("http_proxy");
+    unsetenv("all_proxy");
```
m-blaha: Hm, these variables are handled by curl, right? TBH, rather than changing the global environment, I'd prefer using librepo's standard mechanism for disabling proxies - something like `lr_handle_setopt(handle, NULL, LRO_PROXY, "");` (if you think it's necessary).

prownd: Hmm, that does sound like it.
Modify it then.

m-blaha (approved):
Thank you!

PR #394, opened 2026-09-13
Fix leaks of zchunk contexts on error paths

prownd:
zckCtx is an opaque structure with further allocations, it must be released by zck_free(), not by free():

- init_zck_read() used free() and it didn't release the context at all (and returned FALSE instead of NULL) when zck_init_adv_read() failed,
- prep_zck_header() leaked the context when zck_init_adv_read() failed.


m-blaha (approved):
Thank you!

PR #395, opened 2026-09-13
downloader: Free request headers of interrupted transfers

prownd:
When lr_download() stops the running transfers because of an error (e.g. an interruption by SIGINT or a failure with failfast), it freed their curl handles and files but not the list of extra request headers (target->curl_rqheaders), which was leaked.


m-blaha:
@prownd This one needs rebasing (since it conflicts with d30aba37).

prownd:
Thanks for the heads-up. After d30aba37 the teardown in the lr_download_cleanup path goes through transfer_end(), which already frees transfer->curl_rqheaders, so this leak is fixed upstream and the change becomes empty after rebasing. Closing as superseded by d30aba37.

PR #396, opened 2026-09-13
downloader: Accept file descriptor 0 in lr_download()

prownd:
lr_downloadtarget_new() accepts any non-negative file descriptor but lr_download() asserted that it is greater than zero, so a target using file descriptor 0 (e.g. when stdin is closed) aborted the application.


stewartsmith:
looks good to me. I think I may have introduced this a while ago, and also just found the bug last week.

m-blaha (approved):
Thank you!

PR #398, opened 2026-09-13
util: Handle partial writes in lr_copy_content()

prownd:
write() may write fewer bytes than requested (or fail with EINTR), in which case the rest of the buffer was silently lost and the copy (e.g. of a zchunk file found in the cache) was corrupted.


m-blaha (approved):
Thank you!

PR #399, opened 2026-09-13
checksum: Fix a memory leak and an out-of-bounds read

prownd:
- lr_checksum_fd() leaked the EVP_MD_CTX when lseek(), EVP_DigestUpdate() or EVP_DigestFinal_ex() failed (e.g. for a non-seekable descriptor).
- lr_checksum_clear_cache() walked the list of extended attribute names up to the size returned by the first flistxattr() call instead of the size actually filled in by the second one. If the list shrank in between, strlen() read uninitialized memory beyond the filled part.


m-blaha (approved):
Thank you!

PR #400, opened 2026-09-13
Don't pass negative values to ctype functions

prownd:
The argument of tolower() and isalnum() must be representable as an unsigned char (or be EOF), otherwise the behavior is undefined. A plain char is signed on most platforms, so a non-ASCII byte in a checksum type name or in a URL (e.g. UTF-8) caused undefined behavior (and a crash with some C libraries).


m-blaha (approved):
Thank you!

PR #401, opened 2026-09-13
mirrorlist: Close the duplicated descriptor if fdopen() fails

prownd:
Close the duplicated descriptor if fdopen() fails
Beause  If the failure path does not explicitly call close(fd) on the duplicated descriptor, that descriptor leaks.


ppisar (approved):
Good. The CI failure (dnf4 does not recognize repositories relocated to a new path in Fedora ≥ 45) in unrelated.

PR #403, opened 2026-09-13
Fix swapped function arguments

prownd:
- process_repomd_xml() called lseek(fd, SEEK_SET, 0). It only worked because both SEEK_SET and the intended offset are 0.
- lr_yum_download_remote() logged the revision of repomd.xml in place of the function name and vice versa.


m-blaha (approved):
Thank you!

PR #404, opened 2026-09-13
metadata_downloader: Don't use error messages as format strings

prownd:
lr_metadatatarget_append_error() takes a printf-like format, but error messages were passed to it as the format. The messages contain URLs and paths (e.g. "Download failed: Curl error (37): Couldn't read a file:// file for file:///...") which can contain "%" characters (URL escapes like "%20" or a malicious mirror/metalink URL), so they were interpreted as conversion specifications, reading random memory or writing it with "%n". A metalink URL containing "%s%n" crashed lr_download_metadata().

Pass the messages as an argument of a "%s" format.


m-blaha (approved):
Thank you!

PR #405, opened 2026-09-13
package_downloader: Restore the SIGINT handler on all return paths

prownd:
- lr_download_packages() returned directly when stat() of an existing file failed or when the fastest mirror detection failed, leaving librepo's SIGINT handler installed (Ctrl-C then no longer terminated the application) and leaking the prepared download targets and the list of handles for the fastest mirror detection.
- lr_check_packages() installed its SIGINT handler with an uninitialized struct sigaction: sa_mask was never emptied (so random signals were blocked during the handler) and the rest of the structure contained garbage.
- lr_check_packages() dereferenced the handle of the targets without checking it for NULL, which lr_download_packages() allows.


evan-goode (approved):
LGTM, thanks!

PR #406, opened 2026-09-13
repoutil_yum: Use S_ISDIR() to check for a directory

prownd:
"st_mode & S_IFDIR" is also true for block devices (S_IFBLK) and sockets (S_IFSOCK), so lr_repoutil_yum_parse_repomd() treated them as repository directories.


m-blaha (approved):
Thank you!

PR #407, opened 2026-09-13
repomd: Fix a warning message and a memory leak

prownd:
- A header-checksum element without the type attribute was reported as an "open checksum element".
- The cpeid of a distro tag was leaked when the parsing ended with an error before the end of the distro element (or when the element was repeated without being closed).


fhbash (approved):
LGTM, thank you

PR #412, opened 2026-09-17
Add LR_PACKAGEDOWNLOAD_TRANSIENT to skip persisting checksum cache

stewartsmith:
During a standard "dnf5 install foo", packages are downloaded with keepcache=0 (the default), and are deleted shortly after the transaction. This information never reaches librepo, so it still does an fsync() on the package before writing the xattr with the checksum. This just forces a bunch of IO to persistent storage.

This patch makes it possible to tell librepo that you're downloading a package for transient use, and thus it is okay for it to skip the levels of parnoia needed for ensuring something hits persistent storage correctly.

We add LR_PACKAGEDOWNLOAD_TRANSIENT to LrPackageDownloadFlag, which keeps ABI. Then, we need to thread a persist_checksum_cache flag from lr_download_packages() down to where the decision is made to fsync() and write the xattr is made.

The existing LRO_CHECKSUM is for verify-or-not, not persist-or-not. With LR_PACKAGEDOWNLOAD_TRANSIENT, the checksum is still verified, just not persisted to disk before being deleted.

The public lr_download()/lr_download_target() entry points, and every other caller of the checksum-caching code, are unaffected and keep persisting as before - the new behavior is opt-in per lr_download() batch via the new flag.

Implementation:
- checksum.c: hoist the fsync() inside the caching-enabled guard, so caching=FALSE (already a public, but previously unreachable-from- downloads, parameter of lr_checksum_fd_compare/lr_checksum_fd_cmp) actually skips the fsync rather than paying for it and then not writing the xattr it was protecting.
- downloader.c: make lr_download() a wrapper to lr_download_internal(), which has a new persist_checksum_cache parameter, and give it a new gboolean field on the internal (non-public) LrDownload struct.
- package_downloader.c/.h: add the LR_PACKAGEDOWNLOAD_TRANSIENT flag and compute persist_checksum_cache = !(flags & TRANSIENT) in lr_download_packages().
- python/packagedownloader-py.c, python/__init__.py: expose the flag as an optional third positional/keyword arg (transient=False) on download_packages()

I did some performance benchmarks in EC2 on both Fedora 44 and the AL2027 Preview AMIs. To cover large and small memory configurations, I used:

```
 t4g.nano              c8gn.16xlarge
 --------              -------------
 512MB RAM             128GB RAM
 Graviton 2            Graviton 4
 2vCPUs[1]             64 CPUs
 "up to 5gigabit"[1]   200Gbit
```
[1] these are burstable

Each with 1TB EBS volume, default gp3 settings: 3k IOPs, 125MB/sec

To benchmark, I used some package sets derived from the full dependency closures of a "web-service stack", a "sysadmin utility set", and the full GNOME desktop group from AL2023. Thus, the packages were about constant between the Amazon Linux and Fedora tests.

Introducing the transient flag and using it in dnf5 showed a decent improvement in wall clock time for downloading the packages, a minor reduction in disk IO on XFS but a significant reduction on btrfs:
```
                          wall clock         disk bytes written
                          reduction          reduction
  AL2027 (XFS)            3% - 43% faster    2% - 4% less
  Fedora 44 (btrfs)      46% - 68% faster   12% - 24% less
```
With the fsync() and xattr write (before using the transient flag), the math suggests a per-file overhead of ~125-135 KB on btrfs and a ~9-15 KB one on XFS. This is regardless of package set or download concurrency. The 24% improvement in Fedora with btrfs is with the package set of "sysadmin utilities" as those RPMs are ~325kb on average.

----

Associated `dnf5` pull request to use this flag: https://github.com/rpm-software-management/dnf5/pull/2947


m-blaha (approved):
Thank you, I like it! Let’s keep it open for a while to give other maintainers a chance to chime in.

m-blaha:
The dnf-tests failures are unrelated - see https://github.com/rpm-software-management/ci-dnf-stack/pull/1946 .

kontura (approved):
Looks good to me too.

